PR #753, opened 2016-12-31
Update to Contrib Doucment

atoff:
Includes a hint about VT table migration, and a new section detailing access to the admin panel


NFarrington:
Admin sessions are currently the same as user sessions within the main application. To authenticate with the admin panel, you can first authenticate with the main application, and then navigate to /adm/dashboard (not /adm). PR #755 addresses this and forces authentication via the main application first, so the whole process appears linked again.

With respect to migrations, all module migrations can be run at once using `php artisan module:migrate` (without specifying a specific module). Feel free to submit a new PR for just this change :)

PR #788, opened 2017-01-14
Overlapping texts

### LINKED ISSUE #655: Overlapping Texts
On the user dashboard, the text in all boxes with column layouts has a tendency to overflow into neighboring columns when using a small screen - still intelligible but not particularly pretty.

Noticed in Chrome on a 15" laptop.

atoff:
Added a new CSS class so that columns can, where necessary, break the text to fit within their bounds.

Implemented into the dashboard where needed (Community groups, TeamSpeak registrations and account qualifications).

**Gulp will need to be run on deployment**

Closes #655

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -180,7 +180,7 @@ body, html {
   border-top-right-radius: 0px;
 }
 
-.panel-ukblue .panel-body .row-text-contain{
+.panel-ukblue .panel-body .row-text-contain {
   word-break: break-word;
 }
 
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -180,6 +180,10 @@ body, html {
   border-top-right-radius: 0px;
 }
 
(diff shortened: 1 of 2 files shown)


A-Lawrence:
Thanks for this @atoff - whilst I like the general principle, I think we can strengthen it slightly.  It'd be nice if the left column and right column collapsed to be on top/bottom.  I.e.:

```
CURRENT MEMBERSHIPS: Northern
```

Becomes
```
CURRENT MEMBERSHIPS:
Northern
```

atoff:
How's that?


It's difficult for the ATC/Pilot qualifications, as the size at where the overlapping occurs is between device boundaries for bootstrap's columns

A-Lawrence:
That's really nice.

Can you mosifed the ATC/pilot ones if you changed the overall layout of that section?

A-Lawrence:
Modify* damn auto correct.

atoff:
Think I found a solution that works :)

A-Lawrence:
Thanks @atoff :thumbsup:

PR #795, opened 2017-01-17
Feature - Allow TS names to include network callsign

### LINKED ISSUE #791: Allow TS nickname to include callsign
Allow TS nickname to include ATC callsign in the format:

`Joe Bloggs - EGLL_TWR`

https://community.vatsim.uk/topic/34025-teamspeak-prefixsuffix-when-controlling/

A-Lawrence:
Fixes #791


atoff:
When this goes live, it would probably be best to warn people they need to use their **exact** callsign, especially after a controller change, where someone controlling KK TWR might be EGKK__TWR or EGKK_1_TWR etc., to avoid being kicked (As long as I'm reading the implementation of this right)

morphicsphil:
True, or advise them there's not really any need to change to the regular callsign anyway if they are on a relief callsign.

Does the API offer a way to poke people? Maybe that would be less harsh - but also log who was sent a message? I.e. "controller x logged on as y but TS name is z"

A-Lawrence:
@morphicsphil - good shout.  We'll need to modify the poke/kick scenario for this, as I've bolted it onto the rest they'll be kicked straight away... We really could do with a graceful kick scenario (of 2 warnings or something).  Waiting to hear back from @NFarrington on his review and suggestions for improvement.

atoff:
Closing due to my messing up of the push. Re-opened in #890

PR #798, opened 2017-01-17
WSLogging Verbosity

### LINKED ISSUE #782: Push to wslogging based on verbosity settings
Change the main Command class to only push messages to wslogging based on the level of verbosity specified for the task.

atoff:
Fixes #782 
Checks to see if verbosity is specified


A-Lawrence:
Thanks @atoff :thumbsup:

PR #836, opened 2017-01-28
Added environment based configuration for slack

chrisdoherty4:
- Changes for environment based configuration of slack to prevent spamming of logs - sozza for that. 
- Added some extra stuff to gitignore produced by various Linux commands. 
- Updating composer.lock to have increased minor version bumps - nothing big.


config/slack.php
```diff
@@ -38,7 +38,7 @@
   |
   */
 
-  'username' => 'VATSIM UK Bot',
+  'username' => env("SLACK_USERNAME", "SLACK DEFAULT USERNAME"),
```
atoff: Could this default username be something more like "Default Bot" or something?

.env.example
```diff
@@ -42,6 +42,15 @@ MAIL_FROM_NAME=
...
 MAILGUN_SECRET=
 
+# Slack Webhook Endpoint path to be appended to https://hooks.slack.com/services/.
+SLACK_ENDPOINT=
+
+# Default slack channel for the bot to post in.
+SLACK_CHANNEL=
+
+# Default username to post as
+SLACK_USERNAME=
```
atoff: Perhaps this could be renamed to `SLACK_BOT_USERNAME` to make it clear that this is for the bot.
I'm not sure that this needs a comment.

chrisdoherty4: I would dispute the "no need for comments" in this file entirely. I feel at the very least each configuration block should have a block comment explaining what it's for. At present, it's dependent on the developer reading the CONTRIBUTING.md or actual config files themselves which seems counter productive. It makes more sense, to me, to have comments embedded next to where you set values as opposed to relying on some second file to get the details/consequences of setting a config var. E.g. 
```
# The VATSIM_CERT options are used to connect to the single sign-on system. Each value is 
# unique to a division, and a secret so leave them blank
SSO_CERT=
SSO_KEY=VATSIM_UK
SSO_SECRET=
```
It may seem superfluous but it removes doubt and ambiguity right where they're set. 

Regarding the use of "BOT" my logic was that the purpose of the slack configuration is to hook into slack. Core uses it for the purposes of a BOT (which could be highlighted in the comments) but by defining the vars as "BOT" you're constraining the scope of the environment variable. Again, some may see it as superfluous but in my head there's a clear separation.

atoff: I, personally, feel like it would be a bit overkill to have a comment that a `SLACK_USERNAME` variable relates to the username that the bot would post as? (Surely it should be intuitive?) 
With regards to limiting the scope by adding the `_BOT_`, that variable by definition is only used by our slack 'bot' for the purpose of notification. I could also only ever see it being used as such. We would never need to use core's hook into slack to post as an actual member/person, or for it to be used by a 'human'.

I do, however, agree that we could add a comment to each block, but it would need to be concise, as primarily we want the user to look into the contributing document (which can be updated easier, with more room for fancy formatting) for help with initial configuration. The names of the variables should (hopefully) already provide some context as to their purpose.
Feel free to add a commit updating the contributing document to detail the new variables.

Regarding the CONTRIBUTING.md, it's there to help with that initial configuration, just like any other software would come with a manual, and it is there to explain the in's and outs in detail.

chrisdoherty4: Yes, I agree, it is overkill putting it above the username; more so if I switch to "BOT_USERNAME". 

I don't know what the deal is with changing it (newbie). Do I wait for further comments or just change it?

atoff: All you need to do to update your pull request is change the code, commit it, and then push to your branch. In this case, your local development branch. You will see at the bottom of the pull request, it will show the new commits.

.env.example
```diff
@@ -42,6 +42,15 @@ MAIL_FROM_NAME=
 MAILGUN_DOMAIN=
 MAILGUN_SECRET=
 
+# Slack Webhook Endpoint path to be appended to https://hooks.slack.com/services/.
+SLACK_ENDPOINT=
```
A-Lawrence: I'd prefer the endpoint to be a full URL, rather than something to be appended.

config/slack.php
A-Lawrence: Agreed.

.env.example
A-Lawrence: Additionally, could it be `SLACK_ENDPOINT_URL`? See further comments to see why.

.env.example
```diff
@@ -42,6 +42,15 @@ MAIL_FROM_NAME=
...
+
+# Default slack channel for the bot to post in.
+SLACK_CHANNEL=
```
A-Lawrence: Could it be `SLACK_ENDPOINT_CHANNEL`?

.env.example
A-Lawrence: Could it be `SLACK_ENDPOINT_USERNAME`?

This stops us confusing it with the actual BOT username.

NFarrington:
@chrisdoherty4 Do you intend to update this any further at this time, based on the feedback provided?

NFarrington:
Closed due to inactivity

PR #839, opened 2017-01-30
Update online.blade.php

CalumTowers:
Update column header to more appropriate title


A-Lawrence:
Thanks @CalumTowers 👍

PR #848, opened 2017-02-11
Log the IP from which a token was requested

### LINKED ISSUE #847: Log Request IP for SSO Tokens
Add in the ability for the system to log the IP of the system requesting an SSO token.

atoff:
- Added migration to create a new `request_ip` column in sso_tokens
- Implemented new column in SSO/Security when a new token is generated
- New column that is set when the token is used

Closes #847

### DIFF
--- a/database/migrations/2017_02_10_232137_add_request_ip_to_sso_tokens.php
+++ b/database/migrations/2017_02_10_232137_add_request_ip_to_sso_tokens.php
@@ -14,7 +14,7 @@ class AddRequestIpToSsoTokens extends Migration
     public function up()
     {
       Schema::table('sso_token', function (Blueprint $table) {
-          $table->bigInteger('request_ip');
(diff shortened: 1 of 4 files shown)


database/migrations/2017_02_10_232137_add_request_ip_to_sso_tokens.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('sso_token', function (Blueprint $table) {
+            $table->bigInteger('request_ip')->nullable()->after('account_id');
```
A-Lawrence: Can this be made unsigned, please?

atoff:
Thanks @A-Lawrence, this update has been done.

RE A column to check if the token has been used, we could use the soft deleting trait (and add in the deleted_at column), so that once used the token is deleted?

PR #873, opened 2017-03-10
New email branding

### LINKED ISSUE #799: Email layout - new branding
Update the email layout to the new branding (and a bit of a nicer style).

https://github.com/VATSIM-UK/core/blob/development/resources/views/emails/default.blade.php

@CalumTowers are you able to do this, please?

atoff:
Added a new format/styling/layout for our core emails to give a more web-consistent styling for core emails.
It is fully responsive, and has been optimised for older clients as well as newer clients.

Also changed some links to the new domain.

An example email: 
image
(^ Above was a screenshot taken from Outlook - probably the worst client in terms of HTML support. Looks a bit neater on newer clients)

Closes #799


resources/views/emails/mship/account/welcome.blade.php
```diff
@@ -46,26 +46,28 @@
...
     For both Pilot and ATC training our online system will allow you to book mentoring (once you are enrolled as per the above links):
-    {!! link_to("http://rts.vatsim-uk.co.uk/", "http://rts.vatsim-uk.co.uk/") !!}
+    {!! link_to("https://cts.vatsim.uk/", "https://cts.vatsim.uk/") !!}
 </p>
 
 <p>
-    The UK TeamSpeak Server and Forum - used for voice and written communications for the entire UK community:
-    {!! link_to("http://www.vatsim-uk.co.uk/tsreg/", "http://www.vatsim-uk.co.uk/tsreg/") !!}
+    The UK TeamSpeak Server, Slack and Forum - these are used for voice and written communications for the entire UK community. Feel free to join the community and ask questions and meet others:
+    <li> TeamSpeak & Slack - {!! link_to("https://www.vatsim.uk/tsreg/", "https://www.vatsim.uk/tsreg/") !!}</li>
```
NFarrington: `&` should be `&amp;`

resources/views/emails/messages/plain_text.blade.php
```diff
@@ -0,0 +1,6 @@
+Dear {!! isset($recipientName) ? $recipientName : $recipient->name !!},
+
+{{  strip_tags($body) }}
```
NFarrington: Blade `{{ }}` blocks will be run though `htmlspecialchars`, which should make `strip_tags` redundant

resources/views/emails/messages/plain_text.blade.php
```diff
@@ -0,0 +1,6 @@
+Dear {!! isset($recipientName) ? $recipientName : $recipient->name !!},
```
NFarrington: This should use `{{ }}` to appropriately escape the data

resources/views/emails/messages/plain_text.blade.php
atoff: `htmlspecialchars` will replace the html special characters with their `&___;` counterpart, so it is nesseary to completely remove these before that is run.

NFarrington: I see what you mean, good point. Do the plain text emails still look normal after being processed like this?

atoff: Yes, the tags are gone and it displays nicely, helped by the fact that the new lines in the email bodies are conserved. 99% of the time the user will never see these, it's generally used for excerpts and spam detection

atoff:
These are all done too :)

PR #890, opened 2017-04-04
Allow teamspeak nickname to include callsign

### LINKED ISSUE #791: Allow TS nickname to include callsign

atoff:
Allows users that are connected on the network as an active ATC unit (i.e not OBS), to append their callsign to their TeamSpeak nickname.

It follows this logic:
- Their current callsign is appended their real name, and inserted into their possible names in the format "Joe Bloggs - EGKK_TWR".
- VATSIM network data only updates every ~3 mins, so we give the user a grace period if their nickname contains their real name (Joe Bloggs) but does not match a name possibility.
- After the grace period, they are kicked if, after updating the network data again, they still have an incorrect name
- Users are still kicked instantly if their nickname does not contain their real name.

Closes #791


NFarrington:
@atoff Can you resolve these conflicts please?

PR #913, opened 2017-04-27
Updated navbar/layout to show on mobile (for #436)

### LINKED ISSUE #436: Mobile Menu
In gitlab by @CalumTowers on Aug 2, 2016, 19:09

When viewing Core on a mobile device, there is no menu.
At present, this does not hinder functionality but may do in the future.

enlim:
Navbar now drops down on mobile to show menu. I also adjusted a few gutters to have grids align more nicely (alerts/dashboard content not being aligned, transfer boxes were misaligned, etc.)

Closes #436

### DIFF
--- a/resources/views/mship/management/dashboard.blade.php
+++ b/resources/views/mship/management/dashboard.blade.php
@@ -1,9 +1,9 @@
 @extends('layout')
 
 @section('content')
-    <div class="row no-edge-gutters">
+    <div class="row">
         <div class="col-md-6">
-            <div class="col-md-12">
+            <div class="">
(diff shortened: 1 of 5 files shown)


resources/views/mship/management/dashboard.blade.php
```diff
@@ -1,7 +1,7 @@
 @extends('layout')
 
 @section('content')
-    <div class="row">
+    <div class="row no-edge-gutters">
```
NFarrington: Removing col-md-12 from the child elements instead should resolve the padding issue here

PR #929, opened 2017-05-25
Updated nav bar

### LINKED ISSUE #927: Core User Navigation Overhaul
Currently, the core user nav bar is somewhat broad and lacking features.

1) Page categories/dropdowns are too broad, and not intuitive for their content. Possibly should be split into more categories that are more specific to the function of its pages.

2) There are quite a lot of "two-step" pages, so it can be hard for members to remember where things are, and how to get to a specific place. As we continue to add new pages to the public facing core, it will become even more important that it is easily navigated.

Things that need to be addressed:
- Categories that are too broad
- Duplication of navigation items (e.g TS registration appears in both community and account)
- No indication of logged in account. Ideally, we should turn the login button into a profile tab, and rehouse the profile/account links in there.
- The drop-down menus are a little overwhelming. Perhaps we can turn this into a secondary horizontal nav bar, that slides into the light blue bar under the dark blue main navigation.
- Responsive-ness. The core menu needs to be made responsive

enlim:
- Moved feedback outside of a menu for visibility.
- Added resource drop-down that lists resources for users, I linked to NATS charts but I'm not sure if that should be permanent.
- Community still contains visiting/transferring and emailing users. I kept TS/Slack in there instead of account as having duplicates wasn't necessary.
- Account contains the same as before but moved to the right and has a log out button in the drop down.
- Next to the drop down on the right I added a notification bell, but I'm not sure if that should stay: depends on what you plan on using the notifications for.
- Finally, I added a super basic fade in animation for the items, I can update this in the future for a nicer animation when time allows. Possibly expanding from the centre out as mentioned in the spec.

All the menus are now single column and, I feel, better represented by the title.

I tried to have each section with a horizontal subsection but I found this took up a lot of unneeded space and I couldn't find a design which scaled well on mobile, but I can revisit this if you feel it's important.

Closes #927


resources/views/layout.blade.php
```diff
@@ -55,42 +55,53 @@
...
                     </li>
                 </ul>
 
-                @if(Auth::check())
-                    <ul class="nav navbar-nav navcustom">
+                @if(Auth::check() || Session::has('auth.vatsim-sso'))
+                    {!! Form::open(['route' => 'logout', 'id' => 'logout-form']) !!}
+                    <ul class="nav navbar-nav navcustom navbar-right">
                         <li class="dropdown dropdown-large">
                             <a href="#" class="dropdown-toggle" data-toggle="dropdown">Account <b class="caret"></b></a>
```
atoff: Could we change this so that their name and CID is displayed?
Perhaps like "Alex Toff (1294298)"

resources/views/layout.blade.php
```diff
@@ -55,42 +55,53 @@
...
+
+                <ul class="nav navbar-nav navcustom">
+                    <li class="dropdown dropdown-large">
+                        <a href="#" class="dropdown-toggle" data-toggle="dropdown">Community <b class="caret"></b></a>
+                        <ul class="dropdown-menu dropdown-menu-large row mainmenu_dropdown">
+                            <li class="col-sm-12">
                                 <ul>
+                                    <li class="dropdown-header">Membership</li>
+                                    <li>{!! link_to_route("visiting.landing", "Visit the UK") !!}</li>
+                                    <li>{!! link_to_route("visiting.landing", "Transfer to the UK") !!}</li>
```
atoff: Do you think it would be working combining these into one Visit/Transfer link seeing as they go to the same place?

resources/views/layout.blade.php
```diff
@@ -55,42 +55,53 @@
...
+                @if(Auth::check() || Session::has('auth.vatsim-sso'))
+                    {!! Form::open(['route' => 'logout', 'id' => 'logout-form']) !!}
+                    <ul class="nav navbar-nav navcustom navbar-right">
                         <li class="dropdown dropdown-large">
                             <a href="#" class="dropdown-toggle" data-toggle="dropdown">Account <b class="caret"></b></a>
-                            <ul class="dropdown-menu dropdown-menu-large row mainmenu_dropdown">
-                                <li class="col-sm-6">
+                            <ul class="dropdown-menu dropdown-menu-logout dropdown-menu-large row mainmenu_dropdown">
+                                <li class="col-sm-12">
                                     <ul>
```
atoff: Possibly add a new section in here with basic account details?
e.g Their status (Visiting/Transfering/Division Member/Intl. Member etc) and maybe a short summary of their current ratings (S3 | P1) - would have to think about how we manage multiple Pilot ratings

resources/views/layout.blade.php
enlim: I could do that, this is shown in the breadcrumb bar below the banner though, so I wasn't sure you'd want it repeated.

atoff: I personally think that the name/account would be more useful/noticeable up on the nav bar rather than the breadcrumb bar; what do you think? Thoughts @NFarrington?

resources/views/layout.blade.php
enlim: I've pushed a commit to add some details although it could maybe be restructured to take up a little less space, I can have a redesign of the information there if needed. I also added the name to the drop down as you suggested and removed it from the breadcrumb bar. Let me know if that'll work better!

atoff: Looks nice! One other small thing I have picked up on is that the account/notification bell items seem to get chucked out of the menu sometimes when the name is too long in the small-medium transition phase.

image

atoff: And while we are here, do you think we should implement a link someone on the navigation to core admin for those with the permission to view it?

enlim: My bad, completely forgot to consider longer names. I'll fix that when I'm back after the weekend. I could fit the admin panel in somewhere, I was thinking the resources bar as I'm not sure it would fit anywhere else, unless it should have a separate button. Would that work?

atoff: I think a separate button (perhaps just an icon like the notification bell) would work well

resources/assets/sass/app.scss
```diff
@@ -310,6 +316,9 @@ body, html {
 
 .nav>li>a:focus, .nav>li>a:hover {
   background-color: #00b0f0;
+  -webkit-transition: background-color 0.2s, color 0.2s, -webkit-transform 0.2s;
```
NFarrington: This can probably be removed - it's been supported for years now without a vendor prefix

resources/assets/sass/app.scss
```diff
@@ -310,6 +316,9 @@ body, html {
...
+  transition: background-color 0.2s, color 0.2s, transform 0.2s;
+  -webkit-transition-timing-function: ease-out;
```
NFarrington: This only exists with the vendor prefix - there's no transition-timing-function property listed

Vendor prefix can probably just be removed, as with line 319

resources/views/layout.blade.php
```diff
@@ -55,43 +55,69 @@
...
                         <li class="dropdown dropdown-large">
-                            <a href="#" class="dropdown-toggle" data-toggle="dropdown">Account <b class="caret"></b></a>
-                            <ul class="dropdown-menu dropdown-menu-large row mainmenu_dropdown">
-                                <li class="col-sm-6">
+                            <a href="#" class="dropdown-toggle" data-toggle="dropdown">{{Auth::user()->name.' (' .Auth::user()->id.')'}} <b class="caret"></b></a>
+                            <ul class="dropdown-menu dropdown-menu-logout dropdown-menu-large row mainmenu_dropdown">
+                                <li class="col-sm-12">
                                     <ul>
+                                        <li class="dropdown-header">Details</li>
+                                        <li><a>{{ $_account->status_string }} {{ !is_null($_account->primary_state) ? $_account->primary_state->name : 'unknown state' }} Member</a></li>
```
NFarrington: Not all pages include the $_account variable when the user is authenticated, so this should be Auth::user()

resources/views/layout.blade.php
```diff
@@ -101,29 +129,41 @@
...
+                            <a href="{{ route("mship.notification.list") }}" title="Notifications">
+								@if(Auth::user()->has_unread_notifications)
+									<i class="fa fa-bell unread-notifications"></i>
+								@else
+									<i class="fa fa-bell"></i>
+								@endif
+							</a>
+                        </li>
+                    </ul>
+					@if(Auth::user()->roles[0]["id"]==1)
```
atoff: This should be something like `@if (Auth::user()->hasPermission('adm/dashboard'))`, because most people who have access to the dashboard don't have the role that gives all permissions.

enlim: Alright, cheers. I wasn't sure what the best way to handle this would be.

resources/views/layout.blade.php
```diff
@@ -55,75 +55,112 @@
...
-                            <ul class="dropdown-menu dropdown-menu-large row mainmenu_dropdown">
-                                <li class="col-sm-6">
+                            <a href="#" class="dropdown-toggle hidden-xs hidden-sm" data-toggle="dropdown">{{Auth::user()->name.' (' .Auth::user()->id.')'}} <b class="caret"></b></a>
+							<a href="#" class="dropdown-toggle visible-xs visible-sm" data-toggle="dropdown"><i class="fa fa-sliders"></i> <b class="caret"></b></a>
+                            <ul class="dropdown-menu dropdown-menu-logout dropdown-menu-large row mainmenu_dropdown">
+                                <li class="col-sm-12">
                                     <ul>
-                                        <li class="dropdown-header">Password</li>
-                                        <li>{!! link_to_route('password.change', "Modify") !!}</li>
+                                        <li><a>{{ Auth::user()->status_string }} {{ !is_null(Auth::user()->primary_state) ? Auth::user()->primary_state->name : 'unknown state' }} Member</a></li>
```
NFarrington: I think the division state and ratings can be removed from the account dropdown - what do you think @atoff?

resources/views/layout.blade.php
```diff
@@ -55,75 +55,112 @@
...
-                                    <ul>
-                                        <li class="dropdown-header">Third-Party Services</li>
-                                        <li>{!! link_to_route("teamspeak.new", "TS Registration") !!}</li>
-                                        <li>{!! link_to_route("slack.new", "Slack Registration") !!}</li>
+                                        <li>{!! link_to_route("mship.manage.email.add", "Add Email Address") !!}</li>
+                                        <li>{!! link_to_route("mship.manage.email.assignments", "Email Assignments") !!}</li>
+                                        @if(Session::has('auth.vatsim-sso'))
+                                            <li class="divider"></li>
+                                            <li>
+                                                <a href="{{ route('logout') }}" onclick="event.preventDefault(); document.getElementById('logout-form').submit();">LOG OUT</a>
```
NFarrington: 'LOG OUT' can be lowercase.

resources/views/layout.blade.php
enlim: Okay, I'll remove this. On a similar note, I don't have the ATC rating section showing any instructor ratings. Would you like those added? Not sure I they're necessary but might be helpful.

enlim:
Rebased to clean up the 15 or 16 odd commits.

PR #942, opened 2017-06-19
Feature/fix bad date, closes #936

### LINKED ISSUE #936: Ban "Year" displays as "Hours"
As seen below.
A 10 year ban was issued but shows as hours instead of years.

/adm/mship/account/

abpfgfp

enlim:
* Hour -> year.
* Added check for unreliable dates, throws exception where this occurs.

As far as I can see this should resolve any date issues addressed in #936.

Closes #936


app/helpers.php
```diff
@@ -117,12 +117,17 @@ function human_diff_string(\Carbon\Carbon $ts1, \Carbon\Carbon $ts2, $absolute =
         return 'unknown length';
     }
 
+    if ($ts2 <= \Carbon\Carbon::minValue() || $ts2 >= \Carbon\Carbon::maxValue()
+        || $ts1 <= \Carbon\Carbon::minValue() || $ts1 >= \Carbon\Carbon::maxValue()) {
+        return 'unreliable dates';
```
NFarrington: This check should throw an exception rather than returning a string - this will mean any issues can be fixed further up the stack, rather than pretending there's not an issue.

app/helpers.php
```diff
@@ -117,12 +117,17 @@ function human_diff_string(\Carbon\Carbon $ts1, \Carbon\Carbon $ts2, $absolute =
         return 'unknown length';
     }
 
+    if ($ts2 <= \Carbon\Carbon::minValue() || $ts2 >= \Carbon\Carbon::maxValue()
```
NFarrington: Values can be equal to the min/max, so less than/greater than should be used, rather than lt/gt or equal to.

app/helpers.php
enlim: Is it wise to throw an error here? Doing so would prevent any page from loading since there is no error catching above this function in many cases, I could add a more detailed output to allow the problem to be resolved if needed.

enlim:
Added the changes you suggested. Git conflicted with the change itself and I'm not all that great with it so there is probably a few rebase errors in there, sorry about that.

PR #944, opened 2017-06-20
Update Default Feedback Questions

CalumTowers:
Small changed to questions.


database/migrations/2017_03_31_211222_create_feedback_tables.php
```diff
@@ -139,7 +139,7 @@ public function up()
               'type_id' => '4',
               'form_id' => '1',
               'slug' => 'professional4',
-              'question' => 'They were professional and well delivered',
+              'question' => 'They service they provided was professional and well delivered.',
```
atoff: "They service they provided" should probably be "The service"

database/migrations/2017_03_31_211222_create_feedback_tables.php
```diff
@@ -151,7 +151,7 @@ public function up()
               'type_id' => '4',
               'form_id' => '2',
               'slug' => 'professional4',
-              'question' => 'They were professional and well delivered',
+              'question' => 'They service they provided was professional and well delivered.',
```
atoff: Not sure that this question would make sense for pilots, seeing as they don't provide a service

database/migrations/2017_03_31_211222_create_feedback_tables.php
```diff
@@ -211,7 +211,7 @@ public function up()
               'type_id' => '4',
               'form_id' => '1',
               'slug' => 'enjoyed7',
-              'question' => 'I enjoyed flying/controlling',
+              'question' => 'I enjoyed flying/controlling alongside them.',
```
atoff: Because we now have separate forms, this could be made to be just "controlling", and the next question "flying"

database/migrations/2017_03_31_211222_create_feedback_tables.php
CalumTowers: Typo! Agreed

CalumTowers: Agreed it does not make sense at the moment but as pilot feedback will be disabled on launch and we don't have a question to replace it with, I'd suggest we just deploy with this question "hidden in the back end".

NFarrington: @CalumTowers Update from "They service" to "The service" required

Issue #946, opened 2017-06-23
Instant Ban From TS


CalumTowers:
When a ban is applied on Core (local ban) then we should check the TS for any active connections and apply a short (~30 min) ban to tide over until the "full ban" kicks in.
This will mean suspensions issued on Core for TS misconduct can be dealt with instantly and not continue for 10 mins.

NFarrington:
This should be a kick only, or simply the full ban applied. If the client is only kicked, the daemon will then take action the next time they attempt to connect.

CalumTowers:
I think a kick will probably be the best solution and then let the Deamon pick it up on their next login attempt for the ban.

PR #948, opened 2017-06-23
[WIP/TESTING] Sync bans to forum/ts

### LINKED ISSUE #945: Sync Bans With Forum
Appears to have been started in /app/Listeners/Sync/PushToForum.php
Updates to accounts passed through /app/Console/Commands/SyncCommunity.php

We need to ensure that when an account is suspended/unsuspended in Core (either as a result of a Local Ban or a Network Suspension) that this is passed through to the forum instantly.

Create onBanCreate and onBanRepeal listeners.

atoff:
Simply sync a local ban/network ban to the forum account, and also handle the end of the ban/repeal.

Also added in a listener to run the TeamSpeak manager when a user is banned, thus instantly removing them from the server.

Closes #945 & #946


atoff:
@NFarrington I think this is now ready to be tested on a development environment if able?

NFarrington:
I'll see what can/can't be tested in development and let you know if there are any issues.

Ideally syncing to TeamSpeak would use a new job specifically to kick the user if they are present, rather than running a full sync (which already runs every few minutes)

NFarrington:
The code will be easier to manage and maintain with less duplication if we merge PushBanToForum and PushRepealToForum into the same listener (e.g. SyncBanToForum), which can do a simple check on the $event->ban to see if it's in effect.

It might make sense to have one listener that will sync an account to the forum, rather than just the ban, so that the same listener can be used for when other aspects are updated, e.g. states, and qualifications.

atoff:
@NFarrington I have combined these listeners now; looking forward it would be nice to make use of the Account Touched event and sync like that, but the core functionality that we need is there now, and I will log an issue to look at syncing the whole account when touched.

NFarrington:
The listeners for syncing to different services should remain separate (which is a key reason why it's possible to add multiple listeners to a single event) - it was only the PushBanToForum and PushRepealToForum listeners that should be merged into SyncBanToForum, or something similar. Since the code to ban and unban a forum user is almost identical, that's why it makes sense to have those in one listener.

Similarly, the listener could sync the entire account to the forum, because it's also very similar code, pushing to the same service. The listener should only be concerned with the forum, and use a separate listener for pushing to/syncing with TeamSpeak.

So, just to confirm, there should be one listener for each service, not one listener for each event. It may be that one listener is associated with multiple events.

atoff:
Ah ok that makes a bit more sense - I will look into making the listener/event pairing for Account Touched work in a later PR, because at the moment the Account Touched event doesn't seem to be implemented.

PR #950, opened 2017-06-24
Update password updating to require authentication.

### LINKED ISSUE #939: Setting secondary password on account from command line
In the recent auth refactoring, the `setPassword` method has been slightly changed, and now when you run the setPassword method from a non-logged in instance (such as the `php artisan tinker` facility when creating an account for development on a local system etc) you get an error (Trying to get property of non-object). The error is more cosmetic, as the actual setting of the password occurs before the reference to `Auth::user()`

(App/Models/Mship/Account @ Line 985)

enlim:
Closes #939.

Simple check that the user is authenticated before setting password. Updated contributing document to reflect this.

Also encountered big error where after SSO sign in and before password entry Auth::user() was null but expected. Simple fix to require both Auth::check() and vatsim SSO sign in.


resources/views/layout.blade.php
```diff
@@ -94,7 +94,7 @@
                     </li>
                 </ul>
 
-                @if(Auth::check() || Session::has('auth.vatsim-sso'))
+                @if(Auth::check() && Session::has('auth.vatsim-sso'))
```
NFarrington: This should be `||` instead of `&&` - the logout route clears the session, so if they're partially authenticated (`auth.vatsim-sso`) but not actually logged in yet (`Auth::check()`), they can still log out and reset the session.

app/Models/Mship/Account.php
```diff
@@ -975,6 +976,10 @@ public function calculatePasswordExpiry($temporary = false)
      */
     public function setPassword($password, $temporary = false)
     {
+        if (!\Auth::user()) {
+            // user not authenticated, disallow password change in this manner.
+            throw new AuthenticationException();
```
NFarrington: This function should still reset a user's password even if they're not logged in (it's up to the controller to implement the logic that ensures the action is authorized, rather than the model - the model simply does what it's told).

Instead, line 985 (of the previous version) should implement an `Auth::check()`, e.g. `if ($save && Auth::check() && \Auth::user()->id === $this->id)`

enlim: Okay, will update this.

resources/views/layout.blade.php
enlim: I changed it to `&&` because between SSO authentication and entering your password you are not authenticated but this check will still pass, meaning all calls to `Auth::user()` return null which breaks the page. Perhaps a separate dialogue box to log out if you are only authenticated through Vatsim SSO should be added?

enlim: Made a couple changes, now the main account panel will be displayed if you're fully authenticated, but if only through Vatsim SSO then just a logout button shows. This avoids the call to null that was occurring previously.

NFarrington:
Happy to merge after conflicts resolved.

PR #960, opened 2017-06-26
Member search in ATC/Pilot feedback forms

atoff:
Adds in a search button to allow searching for members. Once clicked, the selected member's CID will be automatically placed into the correct form field.

Worth mentioning that the exact name of the user they want to find is not required, and it will try to find a match from our database.
image
image

Also fix a small grammatical error in the feedback migration

### DIFF
--- a/database/migrations/2017_03_31_211222_create_feedback_tables.php
+++ b/database/migrations/2017_03_31_211222_create_feedback_tables.php
@@ -293,7 +293,7 @@ class CreateFeedbackTables extends Migration
         DB::table('mship_feedback_question_types')->insert([
             [
               'name' => 'userlookup',
-              'code' => '<input class="form-control" name="%1$s" type="text" id="%1$s" value="%2$s" placeholder="Enter the user\'s CID e.g 1234567">',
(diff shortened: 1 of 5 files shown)


CalumTowers:
Nice one @atoff

database/migrations/2017_03_31_211222_create_feedback_tables.php
```diff
@@ -293,7 +293,7 @@ public function up()
         DB::table('mship_feedback_question_types')->insert([
             [
               'name' => 'userlookup',
-              'code' => '<input class="form-control" name="%1$s" type="text" id="%1$s" value="%2$s" placeholder="Enter the Users CID e.g 1234567">',
+              'code' => '<input class="form-control" name="%1$s" type="text" id="%1$s" value="%2$s" placeholder="Enter the user\'s CID e.g 1234567">',
```
NFarrington: This will need a new migration to be effective.

atoff: This has now been done 👍

PR #981, opened 2017-07-20
Add support for additional feedback forms

### LINKED ISSUE #959: Additional Feedback Forms
Allow new forms to be created via ADM for event specific feedback etc etc.
We should also allow forms to be disabled and enabled.

enlim:
Add support for custom feedback forms - closes #959.

Adds an enabled field to forms, which is controlled from the config page on the admin panel. New forms default to disabled. 

All forms will automatically have a user lookup question added to the front of the form - not mentioned during form creation so I will probably go back and add that.

New permission includes adm.mship.feedback.list.* - this avoids requiring a new permission for each custom feedback from, but you can still specify atc and pilot forms since these are defaults. 

Create form button thing:
Image of Create new form button box thing
Emails are created as a contact if they do not exist, the name will strip all characters other than `A-Za-z0-9` and replaces it with a space, and key will be the name with whitespace swapped to `_`'s and in upper case (e.g. email-list@vatsim.uk has `name: Email List` and `key: EMAIL_LIST`)

Save/enable button: (enable becomes red 'Disable' when enabled)
Image of config save/enable form


enlim:
Sorry about the StyleCI changes, I kept missing bits and it's picky. I looked for a client side StyleCI plugin for PhpStorm but couldn't see any, is there such a thing on the go?

NFarrington:
See https://www.jetbrains.com/help/phpstorm/configuring-code-style.html for local code styling. PRs do not need to pass StyleCI, since we can automatically fix the styling after merging.

enlim:
Alright, ta. I was hoping for a server synced type plugin like Sonar but I'll have a play around with the built in style settings today.

app/Http/Controllers/Adm/Mship/Feedback.php
```diff
@@ -12,6 +13,44 @@
...
+        $new_question = new Question();
+
+        return $this->viewMake('adm.mship.feedback.new')
+            ->with('question_types', $question_types)
+            ->with('new_question', $new_question);
+    }
+
+    public function postNewForm(UpdateFeedbackFormRequest $request)
+    {
+        $new_ident = isset($_POST['ident']) ? $_POST['ident'] : null;
```
atoff: Ideally should be using `$request->input('ident')` here, and same for line below

app/Http/Controllers/Adm/Mship/Feedback.php
```diff
@@ -12,6 +13,44 @@
...
+        $new_name = isset($_POST['name']) ? $_POST['name'] : null;
+        $new_contact = isset($_POST['contact']) ? $_POST['contact'] : null;
+        if ($new_ident == null) {
+            return Redirect::back()
+                ->withInput($request->input())
```
atoff: Could some of this validation code be moved into the UpdateFeedbackFormRequest, or a new Request class made for it? You can process the setting of new name/contact in the `getValidatorInstance()` of the request

resources/views/adm/mship/feedback/new.blade.php
```diff
@@ -0,0 +1,173 @@
+@extends('adm.layout')
+
+@section('scripts')
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-sortable/0.9.13/jquery-sortable-min.js"
+            integrity="sha384-mwD0+87SDVjJjyfTMQHNVV+IyWDM38MhzdCFZ+SRefmD75v+M5K0R3naFNLnZf1L"
+            crossorigin="anonymous"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-datetimepicker/2.5.4/build/jquery.datetimepicker.full.min.js" integrity="sha384-8Lb23xW0dVl+HHrv90JF6PpwygXa7Z0zZIK9+RWorNDyubrG7Ppu7JJw32U8op0i" crossorigin="anonymous"></script>
```
atoff: Datetime picker is (should be) already published into public assets by gulp, so we should reference that instead. Likewise, so is sortable

enlim:
I initially had no CID question at the start, however the feedback system currently uses it to associate any feedback with an account - I could look into removing this association for these forms?

atoff:
Either that, or factor into the system that that field may not be present and handle that appropriately, because eventually they will want to use the system for event feedback etc. that doesn't have or need a subject of the feedback (Prehaps implementing feedback types? - or do you think it should be left for another issue later?)

resources/views/adm/mship/feedback/new.blade.php
enlim: Seems that they're not unfortunately, browser can't find them once they're removed and I remade the assets.

enlim:
Alright I'll see what I can do with that.

resources/views/adm/mship/feedback/new.blade.php
atoff: Looks like they have been removed since the move to webpack - if you want to add them into the pack go for it. The form configuration page has been broken as a result of this aswell, so that will need to be fixed

enlim: I had a brief look at this but from what I've seen you can't import JS from a URL using webpack unless you use third party modules, but since these modules aren't needed much I'd argue that it's simpler to just import them using script tags where needed. Please correct me if there is a simpler way to import external JS files with webpack though.

NFarrington: The use of JS plugins such as datepicker still needs a bit more work due to the fact that they were previously used from AdminLTE, which we need to move to a more central location.

Using a CDN URL is fine for now, so feel free to stick with that until we can sort something better out

enlim: Will leave as is then, ta.

NFarrington:
@atoff Can you approve the changes you requested if you're happy, please?

@enlim As far as you're concerned, is this now ready to test/merge?

enlim:
All fine for me, I can submit a fix for the StyleCI changes later this evening and I'll update my branch with that.

PR #985, opened 2017-08-05
Update V/T Submission Statement

CalumTowers:
As per request from BJ


CalumTowers:
Also some random StyleCI updates...

PR #1003, opened 2017-08-31
Stop Banning Users on Moodle

### LINKED ISSUE #1001: Moodle Bans
We should stop syncing bans with Moodle as it is a training function and a local ban should not restrict this.

For network suspensions, users should be suspended from Moodle.

CalumTowers:
Resolves #1001


CalumTowers:
@NFarrington Can we implement this as a hotfix please

PR #1047, opened 2017-12-14
Teamspeak: Default channel idle time

### LINKED ISSUE #1046: Default Channel Idle Timeout
Kick users idle for more than 10 minutes in channel id 1

atoff:
Closes #1046

### DIFF
--- a/app/Libraries/TeamSpeak.php
+++ b/app/Libraries/TeamSpeak.php
@@ -15,6 +15,7 @@ use DB;
 use TeamSpeak3;
 use TeamSpeak3_Adapter_ServerQuery_Exception;
 use TeamSpeak3_Node_Client;
+use TeamSpeak3_Node_Server;
 
 /**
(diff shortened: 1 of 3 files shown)


app/Libraries/TeamSpeak.php
```diff
@@ -359,6 +360,11 @@ public static function checkClientIdleTime(TeamSpeak3_Node_Client $client, Accou
             $maxIdleTime = 60;
         }
 
+        if($server->channelGetById($client->cid)->getId() == 1){
```
NFarrington: The client's channel ID can be accessed via `$client['cid']`

NFarrington: The channel ID should be the default channel ID, rather than having `1` hardcoded. The default channel can be obtained using `$client->getParent()->channelList(['channel_flag_default' => 1])`

Issue #1070, opened 2017-12-27
Referee follow up emails not deleted once submitted


atoff:
When a application is submitted, we add a delayed reference request email.

However, if they complete the application before this follow up email, they still get sent the follow up email

NFarrington:
Duplicate of #1010

PR #1071, opened 2017-12-27
V/T Bug Fixes

### LINKED ISSUE #1010: ApplicationReferenceRequest is still sent after the reference has been submitted
A second ApplicationReferenceRequest notification is queued to be sent 7 days after the first. If the reference has already been submitted, it should not be sent.

The logic for this has been defined in the 'via' method in `/app/Notifications/ApplicationReferenceRequest.php`, which is called when queuing the notifications, instead of when the notification is ready to be sent.

This logic should be refactored so that the visit/transfer cron or a queued job (instead of queued notification) makes the check and then sends the reminder notification if the reference hasn't been submitted.

atoff:
1. Addresses #1070/#1010, implements a listener to remove reminder notifications that are no longer needed
2. Add a change to facility create/update validation rules, and change the database structure so that infinity training spaces actually works

Closes #1010


app/Listeners/VisitTransfer/NotifyAllReferees.php
```diff
@@ -17,22 +21,34 @@ public function __construct()
     public function handle(ApplicationSubmitted $event)
     {
         $refereesToBeNotified = $event->application->referees->filter(function ($ref) {
-            return !$ref->is_requested && !$ref->is_submitted;
+            if($ref->reminded_at == null){
+                return !$ref->is_submitted;
+            }
+            return $ref->reminded_at->lte(Carbon::now()) && !$ref->is_submitted;
```
NFarrington: `->lte(Carbon::now())` should be `->isPast()`

NFarrington: As far as I can tell, this listener should be fired once and only once, after an application has been submitted - if so, is there ever a situation where `reminded_at` would be set at this point?

atoff: This listener is set later to be called again for the second time for the follow up emails, so references that are not submitted will have reminded at set, which would cause an error in the following line

Issue #1081, opened 2018-01-04
Implement Complete Button


CalumTowers:
Add "Complete Application" button alongside Reject and Accept for the application view page

atoff:
Resolved in #1082

PR #1082, opened 2018-01-04
Implement Complete Application Button

CalumTowers:
Implement a button to manually complete applications

### DIFF
--- a/app/Http/Controllers/VisitTransfer/Admin/Application.php
+++ b/app/Http/Controllers/VisitTransfer/Admin/Application.php
@@ -6,6 +6,7 @@ use App\Http\Controllers\Adm\AdmController;
 use App\Http\Requests\VisitTransfer\ApplicationAcceptRequest;
 use App\Http\Requests\VisitTransfer\ApplicationCheckOutcomeRequest;
 use App\Http\Requests\VisitTransfer\ApplicationRejectRequest;
+use App\Http\Requests\VisitTransfer\ApplicationCompleteRequest;
 use App\Http\Requests\VisitTransfer\ApplicationSettingToggleRequest;
 use App\Models\Mship\Account;
 use App\Models\VisitTransfer\Application as ApplicationModel;
@@ -110,6 +111,19 @@ class Application extends AdmController
         return Redirect::back()
                        ->withSuccess('Application #'.$application->public_id.' - '.$application->account->name.' accepted &amp; candidate notified.');
     }
+
+    public function postComplete(ApplicationCompleteRequest $request, ApplicationModel $application)
+    {
+        try {
+            $application->complete(Input::get('complete_staff_note', null), Auth::user());
+        } catch (\Exception $e) {
+            return Redirect::back()->withError($e->getMessage());
+        }
+
+        return Redirect::back()
+            ->withSuccess('Application #'.$application->public_id.' - '.$application->account->name.' completed.');
+    }
+
 
     public function postCheckMet(ApplicationCheckOutcomeRequest $request, ApplicationModel $application)
     {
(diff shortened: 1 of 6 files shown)


CalumTowers:
Resolves #1082

CalumTowers:
Resolves #1081

app/Http/Requests/VisitTransfer/ApplicationCompleteRequest.php
```diff
@@ -0,0 +1,47 @@
...
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
+        $application = $this->route('application');
+
+        return Gate::allows('complete', $application);
+        //return true;
```
atoff: left over?

app/Http/Requests/VisitTransfer/ApplicationCompleteRequest.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'complete_staff_note' => 'nullable|string',
```
atoff: Do we want/need a length limit here? Because any single letter on its own would suffice using this validation. e.g `min:25`

resources/views/visit-transfer/admin/application/view.blade.php
```diff
@@ -545,7 +559,47 @@ class="text-danger">This action cannot be undone</strong>.
...
+                        </h4>
+                    </div>
+                    <div class="modal-body">
+                        <p>
+                            It is important that you understand <strong class="text-danger">this action cannot be
+                                undone.</strong>
+                        </p>
+
+                        <p>
+                            You must write a staff note detailing why you have completed the application.
```
atoff: You say here that you must write a staff note, however you allow the note to be nullable in the validation

app/Http/Controllers/VisitTransfer/Admin/Application.php
```diff
@@ -111,6 +112,18 @@ public function postAccept(ApplicationAcceptRequest $request, ApplicationModel $
                        ->withSuccess('Application #'.$application->public_id.' - '.$application->account->name.' accepted &amp; candidate notified.');
     }
 
+    public function postComplete(ApplicationCompleteRequest $request, ApplicationModel $application)
+    {
+        try {
+            $application->complete(Input::get('complete_staff_note', null), Auth::user());
+        } catch (\Exception $e) {
+            return Redirect::back()->withError($e->getMessage());
```
atoff: This blanket try catch for all Exceptions would stop us from being able to log any errors as a result of the complete method. Do we even need the try catch at all? (Also if we have a database error or likewise it may potentially give the user information they are not meant to see)

CalumTowers: I've implemented this how its been done for Reject and Accept, if we want to change this I would suggest changing it for the others too?

resources/views/visit-transfer/admin/application/view.blade.php
CalumTowers: Now ensuring that it is mandatory

app/Http/Controllers/VisitTransfer/Admin/Application.php
atoff: I would say so. Thoughts @NFarrington ?

NFarrington: The inconsistency in how we implement the methods that throw these exceptions means that more work is required than the scope of this PR - as the redirect provides the exception message to the user it's not a massive deal.

However, it should be easy to change the blanket Exception to the actual exceptions that need catching, rather than all exceptions, so that should be done for all 3 in this PR.

PR #1093, opened 2018-01-10
Tawk.To on prod only

CalumTowers:
Only show Tawk.To chat script if on production

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -272,6 +272,7 @@
 
 </script>
 
+@if(env('APP_ENV') == 'production')
 <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


resources/views/layout.blade.php
```diff
@@ -272,6 +272,7 @@ class="fa fa-sliders"></i> <b class="caret"></b></a>
 
 </script>
 
+@if(env('APP_ENV') == 'production')
```
NFarrington: should be: `@if(App::environment('production'))`

CalumTowers:
@NFarrington Its to avoid confusion for agents on Tawk.To. Lets keep it to prod only.

PR #1112, opened 2018-02-03
Implement local bans view page

### LINKED ISSUE #1108: List of all Bans
Create page to list all bans, paginate and order by created
Page is permission controlled, can modify ban from this page too

CalumTowers:
Add a page that shows all local bans on one page

Resolved #1108

### DIFF
--- a/app/Http/Controllers/Adm/Mship/Account.php
+++ b/app/Http/Controllers/Adm/Mship/Account.php
@@ -363,6 +363,7 @@ class Account extends AdmController
         $ban->repeal();
 
         $ban->account->notify(new BanRepealed($ban));
+        
         return Redirect::route('adm.mship.account.details', [$ban->account_id, 'bans', $ban->id])
             ->withSuccess('Ban has been repealed.');
     }
@@ -394,6 +395,7 @@ class Account extends AdmController
             Auth::getUser()
         );
         $ban->notes()->save($note);
+
         return Redirect::route('adm.mship.account.details', [$ban->account_id, 'bans', $ban->id])
             ->withSuccess('Your comment for this ban has been noted.');
     }
@@ -446,6 +448,7 @@ class Account extends AdmController
         $ban->save();
 
         $ban->account->notify(new BanModified($ban));
+
         return Redirect::route('adm.mship.account.details', [$ban->account_id, 'bans', $ban->id])
             ->withSuccess('This ban has been modified.');
     }
--- a/app/Http/Controllers/Adm/Mship/Account.php
+++ b/app/Http/Controllers/Adm/Mship/Account.php
@@ -333,8 +333,7 @@ class Account extends AdmController
             ->paginate(15);
 
         return $this->viewMake('adm.mship.account.ban.index')
-            ->with('bans', $bans)
-            ->with('account', $account);
+            ->with('bans', $bans);
     }
 
     public function getBanRepeal(AccountData\Ban $ban)
--- a/app/Http/Controllers/Adm/Mship/Account.php
+++ b/app/Http/Controllers/Adm/Mship/Account.php
@@ -326,7 +326,7 @@ class Account extends AdmController
             ->withSuccess('You have successfully banned this member.');
     }
 
-    public function getBans(AccountData $account)
+    public function getBans()
     {
         $bans = BanData::isLocal()
             ->orderByDesc('created_at')
--- a/app/Http/Controllers/Adm/Mship/Account.php
+++ b/app/Http/Controllers/Adm/Mship/Account.php
@@ -363,7 +363,7 @@ class Account extends AdmController
         $ban->repeal();
 
         $ban->account->notify(new BanRepealed($ban));
-        
+
         return Redirect::route('adm.mship.account.details', [$ban->account_id, 'bans', $ban->id])
             ->withSuccess('Ban has been repealed.');
     }
--- a/app/Http/Controllers/Adm/Mship/Account.php
+++ b/app/Http/Controllers/Adm/Mship/Account.php
@@ -8,6 +8,7 @@ use App\Http\Requests\Mship\Account\Ban\CreateRequest;
 use App\Http\Requests\Mship\Account\Ban\ModifyRequest;
 use App\Http\Requests\Mship\Account\Ban\RepealRequest;
 use App\Models\Mship\Account as AccountData;
+use App\Models\Mship\Account\Ban as BanData;
 use App\Models\Mship\Ban\Reason;
 use App\Models\Mship\Note\Type;
 use App\Models\Mship\Note\Type as NoteTypeData;
@@ -323,6 +324,17 @@ class Account extends AdmController
 
         return Redirect::route('adm.mship.account.details', [$account->id, 'bans', $ban->id])
             ->withSuccess('You have successfully banned this member.');
+    }
+
+    public function getBans(AccountData $account)
+    {
+        $bans = BanData::orderBy('created_at', 'DESC')
+            ->where('type', '80') //Only Show Local Bans
+            ->paginate(15);
+
+        return $this->viewMake('adm.mship.account.ban.index')
+            ->with('bans', $bans)
+            ->with('account', $account);
     }
 
     public function getBanRepeal(AccountData\Ban $ban)
--- a/app/Http/Controllers/Adm/Mship/Account.php
+++ b/app/Http/Controllers/Adm/Mship/Account.php
@@ -329,7 +329,7 @@ class Account extends AdmController
     public function getBans(AccountData $account)
     {
         $bans = BanData::isLocal()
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Adm/Mship/Account.php
```diff
@@ -325,6 +326,17 @@ public function postBanAdd(CreateRequest $request, AccountData $account)
             ->withSuccess('You have successfully banned this member.');
     }
 
+    public function getBans(AccountData $account)
+    {
+        $bans = BanData::orderBy('created_at', 'DESC')
+            ->where('type', '80') //Only Show Local Bans
```
atoff: Should use BanData::TYPE_LOCAL (just in case we change it or something) instead of the 80. If your feeling up to it, you could even implement this as a scope in the Account\Ban model (:)))

resources/views/adm/mship/account/ban/index.blade.php
```diff
@@ -0,0 +1,133 @@
...
+            </div><!-- /.box -->
+        </div>
+    </div>
+@stop
+
+@section('scripts')
+    @parent
+    {!! HTML::script('/AdminLTE/js/plugins/datatables/jquery.dataTables.js') !!}
+    {!! HTML::script('/AdminLTE/js/plugins/datatables/dataTables.bootstrap.js') !!}
+    <script type="text/javascript">
```
atoff: Merge these two script blocks into one

app/Http/Controllers/Adm/Mship/Account.php
CalumTowers: @atoff The fact that this scope was already there and I didn't use it... makes me sad... Sorted!

app/Http/Controllers/Adm/Mship/Account.php
```diff
@@ -325,6 +326,17 @@ public function postBanAdd(CreateRequest $request, AccountData $account)
             ->withSuccess('You have successfully banned this member.');
     }
 
+    public function getBans(AccountData $account)
+    {
+        $bans = BanData::isLocal()
+            ->orderBy('created_at', 'DESC')
```
NFarrington: `->orderByDesc('created_at')`

app/Http/Controllers/Adm/Mship/Account.php
```diff
@@ -352,8 +364,9 @@ public function postBanRepeal(RepealRequest $request, AccountData\Ban $ban)
 
         $ban->account->notify(new BanRepealed($ban));
 
-        return Redirect::route('adm.mship.account.details', [$ban->account_id, 'bans', $ban->id])
-            ->withSuccess('Ban has been repealed.');
+        $return = Input::get('return');
```
NFarrington: `$return = $request->input('return');`

app/Http/Controllers/Adm/Mship/Account.php
```diff
@@ -384,8 +397,9 @@ public function postBanComment(CommentRequest $request, AccountData\Ban $ban)
         );
         $ban->notes()->save($note);
 
-        return Redirect::route('adm.mship.account.details', [$ban->account_id, 'bans', $ban->id])
-            ->withSuccess('Your comment for this ban has been noted.');
+        $return = Input::get('return');
```
NFarrington: `$return = $request->input('return');`

app/Http/Controllers/Adm/Mship/Account.php
```diff
@@ -437,8 +451,9 @@ public function postBanModify(ModifyRequest $request, AccountData\Ban $ban)
 
         $ban->account->notify(new BanModified($ban));
 
-        return Redirect::route('adm.mship.account.details', [$ban->account_id, 'bans', $ban->id])
-            ->withSuccess('Your comment for this ban has been noted.');
+        $return = Input::get('return');
```
NFarrington: `$return = $request->input('return');`

resources/views/adm/layout/sidebar.blade.php
```diff
@@ -92,6 +92,12 @@
                                             class="fa fa-angle-double-right"></i> Accounts List</a>
                             </li>
                         @endif
+                        @if($_account->hasChildPermission("adm/mship/bans"))
```
NFarrington: Single quotes

resources/views/adm/layout/sidebar.blade.php
```diff
@@ -92,6 +92,12 @@
...
+                            <li {!! (\Request::is('adm/mship/bans*') ? ' class="active"' : '') !!}>
+                                <a href="{{ URL::route("adm.mship.ban.index") }}"><i
```
NFarrington: Single quotes (inside function call)

CalumTowers:
@NFarrington How likely is it that someone will be going directly to a modification, repeal or notes page for a Ban?

If we really don't want to do it, I think we should probably just redirect back to the Adm Dashboard with a success message - generally no further action is required on the user's account after finishing and going to a list of all bans would not be very useful.

As for the sidebar and the single quotes, my code is following the convention of the rest of the sidebar. Would you like the others amending within this PR too?

NFarrington:
It's unlikely, but possible, and if there's a possibility of something happening, that needs to be handled appropriately. It could be that a member of the DSG links the form to another member to fill out. It could be they bookmark/save it and come back to it later.

I definitely think we should be avoiding the use of `previous()` in these situations, since it doesn't add much value and is likely to cause problems rather than solve them.

You could redirect to a user's account after updating the ban, which would semantically likely make more sense to the user. A lot of the design decisions end up being to do with HCI and how the user is actually using the system.

With the sidebar and single quotes, not necessarily a big deal - if others on the same page also need changing, they can be left out for now

CalumTowers:
@NFarrington I've reverted my amends to redirection so it will just go back to the banned user's account page on the ban tab.

I've left the sidebar as is.

resources/views/adm/mship/account/ban/repeal.blade.php
```diff
@@ -37,7 +37,7 @@
 
                     <div class="btn-toolbar">
                         <div class="btn-group pull-left">
-                            {!! link_to_route("adm.mship.account.details", "Cancel", [$ban->account_id, "bans", $ban->id], ["class" => "btn btn-danger"]) !!}
+                            <a href="{{ url()->previous() }}" class="btn btn-danger">Cancel</a>
```
NFarrington: These should also be linked statically.

resources/views/adm/mship/account/ban/modify.blade.php
```diff
@@ -57,7 +57,7 @@
 
                     <div class="btn-toolbar">
                         <div class="btn-group pull-left">
-                            {!! link_to_route("adm.mship.account.details", "Cancel", [$ban->account_id, "bans", $ban->id], ["class" => "btn btn-info"]) !!}
+                            <a href="{{ url()->previous() }}" class="btn btn-danger">Cancel</a>
```
NFarrington: These should also be linked statically.

app/Http/Controllers/Adm/Mship/Account.php
```diff
@@ -325,6 +326,17 @@ public function postBanAdd(CreateRequest $request, AccountData $account)
...
 
+    public function getBans(AccountData $account)
+    {
+        $bans = BanData::isLocal()
+            ->orderByDesc('created_at')
+            ->paginate(15);
+
+        return $this->viewMake('adm.mship.account.ban.index')
+            ->with('bans', $bans)
+            ->with('account', $account);
```
NFarrington: You're adding `$account` here but seem to be using the existing `$_account` variable in the view. This line can probably be removed?

resources/views/adm/mship/account/ban/comment.blade.php
```diff
@@ -27,7 +27,7 @@
 
                     <div class="btn-toolbar">
                         <div class="btn-group pull-left">
-                            {!! link_to_route("adm.mship.account.details", "Cancel", [$ban->account_id, "bans", $ban->id], ["class" => "btn btn-info"]) !!}
+                            <a href="{{ url()->previous() }}" class="btn btn-danger">Cancel</a>
```
NFarrington: These should also be linked statically.

resources/views/adm/mship/account/ban/index.blade.php
```diff
@@ -0,0 +1,125 @@
...
+                            @if($_account->hasPermission("adm/mship/ban/*/repeal"))
+                                <th>Repeal</th>
+                            @endif
+                        </tr>
+                        </thead>
+                        <tbody>
+                        @foreach($bans as $b)
+                            <tr>
+                                <td>
+                                    {!! link_to_route("adm.mship.account.details", $b->account->name, [$b->account]) !!}
```
NFarrington: Since this is a brand new file, should definitely be using single quotes for function arguments throughout, unless necessary for technical reasons).

PR #1113, opened 2018-02-04
Gatwick endorsement checker

CalumTowers:
Allow members to easily check their progress to meeting the requirements to apply for a Gatwick endorsement. Requirements listed here: https://vatsim.uk/gatwick-endorsement/


app/Http/Controllers/Controllers/GatwickController.php
```diff
@@ -0,0 +1,73 @@
...
+            ->where('type', 'atc')
+            ->orderBy('created_at', 'DESC')
+            ->first();
+
+        $s1 = $rating->name_long == 'Student 1';
+
+        if ($rating->name_long == 'Observer') {
+            return Redirect::back()
+                ->withError('Only S1 rated controllers are eligible for a Gatwick Ground endorsement.');
+        } elseif ($rating->name_long == 'Student 1') {
```
atoff: This should be of the form:
Else if(!$s1){
// Redirect
}

// Make view here outside if statement

app/Http/Controllers/Controllers/GatwickController.php
```diff
@@ -0,0 +1,73 @@
...
+            return $this->viewMake('controllers.gatwick')
+                ->with('account', $account)
+                ->with('groupone', $g1)
+                ->with('grouptwo', $g2)
+                ->with('groupthree', $g3)
+                ->with('divisionmember', $divisionmember)
+                ->with('s1', $s1);
```
atoff: What is the point of passing S1??

app/Http/Controllers/Controllers/GatwickController.php
```diff
@@ -0,0 +1,73 @@
...
+            ->first();
+
+        $s1 = $rating->name_long == 'Student 1';
+
+        if ($rating->name_long == 'Observer') {
+            return Redirect::back()
+                ->withError('Only S1 rated controllers are eligible for a Gatwick Ground endorsement.');
+        } elseif ($rating->name_long == 'Student 1') {
+            return $this->viewMake('controllers.gatwick')
+                ->with('account', $account)
```
atoff: You shouldn't need to pass the account, it should already be accessible without passing it

resources/views/controllers/gatwick.blade.php
```diff
@@ -0,0 +1,133 @@
...
+                        <li>Stansted (EGSS)</li>
+                        <li>Liverpool (EGGP)</li>
+                    </ul>
+                    <div class="progress">
+                        @if($groupone == 0)
+                            <div class="progress-bar" role="progressbar" style="width: 0%" aria-valuemin="0" aria-valuemax="10"></div>
+                        @elseif($groupone > 10)
+                            <div class="progress-bar progress-bar-success" role="progressbar" style="width: 100%" aria-valuemin="0" aria-valuemax="10">{{ $groupone }} Hours</div>
+                        @else
+                            <div class="progress-bar" role="progressbar" style="width: {{ $groupone*10 }}%" aria-valuemin="0" aria-valuemax="10">{{ $groupone }} Hours</div>
```
atoff: What is the point of checking if they have 0 hours? The progress pass will still be displayed with zero hours exactly the same way. (0*10=0)

resources/views/controllers/gatwick.blade.php
```diff
@@ -0,0 +1,133 @@
...
+                        <li>Luton (EGGW)</li>
+                    </ul>
+                    <div class="progress">
+                        @if($grouptwo == 0)
+                            <div class="progress-bar" role="progressbar" style="width: 0%;" aria-valuemin="0" aria-valuemax="10"></div>
+                        @elseif($grouptwo > 10)
+                            <div class="progress-bar progress-bar-success" role="progressbar" style="width: 100%" aria-valuemin="0" aria-valuemax="10">{{ $grouptwo }} Hours</div>
+                        @else
+                            <div class="progress-bar" role="progressbar" style="width: {{ $grouptwo*10 }}%" aria-valuemin="0" aria-valuemax="10">{{ $grouptwo }} Hours</div>
+                        @endif
```
atoff: Same as above

resources/views/controllers/gatwick.blade.php
```diff
@@ -0,0 +1,133 @@
...
+                        <li>Newcastle (EGNT)</li>
+                        <li>East Midlands (EGNX)</li>
+                    </ul>
+                    <div class="progress">
+                        @if($groupthree == 0)
+                            <div class="progress-bar" role="progressbar" style="width: 0%" aria-valuenow="0" aria-valuemin="0" aria-valuemax="5"></div>
+                        @elseif($groupthree > 5)
+                            <div class="progress-bar progress-bar-success" role="progressbar" style="width: 100%" aria-valuemin="0" aria-valuemax="5">{{ $groupthree }} Hours</div>
+                        @else
+                            <div class="progress-bar" role="progressbar" style="width: {{ $groupthree*20 }}%" aria-valuemin="0" aria-valuemax="5">{{ $groupthree }} Hours</div>
```
atoff: Same as above

resources/views/controllers/gatwick.blade.php
```diff
@@ -0,0 +1,133 @@
...
+            </div>
+        </div>
+        <div class="col-md-4">
+            <div class="panel panel-ukblue">
+                <div class="panel-heading"><i class="glyphicon glyphicon-info-sign"></i> &thinsp; Request Moodle Course</div>
+                <div class="panel-body">
+                    Once you have completed the requirements above, you will be able to press the button below to request access to the Moodle course and progress to Step 2.
+                    <br><br>
+                    @if($divisionmember == true && $groupthree > 5 && $grouptwo > 10 && $groupone > 10)
+                        <a href="mailto:atc-training@vatsim-uk.co.uk?Subject=Gatwick%20Endorsement%20-%20Moodle%20Request&Body=Please%20grant%20me%20access%20to%20the%20Gatwick%20Endorsement%20Moodle%20course%20as%20I%20have%20now%20met%20the%20number%20of%20hours%20required%20across%20the%20three%20groups.%0A%0AGroup%201%3A%20{{ $groupone }} hours.%0AGroup%202%3A%20{{ $grouptwo }} hours.%0AGroup%203%3A%20{{ $groupthree }} hours.%0A%0AFull%20Name%3A%20{{ $account->name }}%0AVATSIM%20CID%3A%20{{ $account->id }}">
```
atoff: Get rid of the == true, it is not needed

app/Http/Controllers/Controllers/GatwickController.php
```diff
@@ -0,0 +1,73 @@
...
+class GatwickController extends \App\Http\Controllers\BaseController
+{
+    public function getIndex()
+    {
+        $account = $this->account;
+
+        $groupone = $this->account->networkDataAtc()
+            ->whereBetween('connected_at', [Carbon::now()->subMonth(3), Carbon::now()])
+            ->where(function ($callsigns) {
+                $callsigns->where('callsign', 'like', 'EGCC_%')
```
atoff: Could this also be refactored into a scope? 
  public static function scopeWithCallsign($query, $callsign)
    {
        return $query->where('callsign', 'LIKE', $callsign);
    }

we could make a new one of these to accept an array of callsigns, somthing like
public static function scopeWithManyCallsigns($query, array $callsigns)
    {
        $query->where('callsign', 'LIKE', $callsigns[0]);
        $callsigns = array_shift($callsigns)
        foreach($callsigns as $callsign){
            $query->orWhere('callsign', 'LIKE', $callsign);
         }
        return $query;
    }

CalumTowers:
@atoff All comments actioned, can you take another look please

app/Http/Controllers/Controllers/EndorsementController.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace App\Http\Controllers\Controllers;
```
NFarrington: This needs moving to a more appropriate namespace.

CalumTowers:
@NFarrington I've renamed the controller to avoid confusion but have left the routes as "Controllers" as it will be more in line with "Pilots" when implemented via smartCars.

Happy to generalise more for use elsewhere - how would you like to see the requirements/groups/ratings implemented for use elsewhere?

CalumTowers:
@NFarrington or @atoff Any support/help/advice to tidy up the `gatwick_ground` view would be fab

app/Http/Controllers/Atc/EndorsementController.php
```diff
@@ -0,0 +1,37 @@
...
+use DB;
+
+class EndorsementController extends \App\Http\Controllers\BaseController
+{
+    public function getGatwickGroundIndex()
+    {
+        $requirements = DB::table('endorsements')->where('endorsement', '=', 'EGKK_GND')->get();
+        $outcomes = collect([]);
+
+        foreach ($requirements as $r) {
```
NFarrington: This can be made cleaner by calling `transform()` or `map()` on `$requirements`

NFarrington:
For the view, at first glance you should be able to do `@foreach($outcomes as $outcome)` and loop through each outcome for each of the sections.

In the controller, when mapping or transforming, it might make more sense to use a stdClass object with two properties, the requirement and the outcome, rather than losing meaning in an array.

CalumTowers:
@NFarrington If we do a foreach on the outcomes, the text will need to be stored in the database too then I guess for each box?

Concerned this will over complicate massively.

NFarrington:
@CalumTowers Is the text for each group not the same essentially, with a few variables? For the airport ICAOs, they're already in the DB for the requirements, and if you really wanted you could pull the names from the `airports` table.

Shouldn't matter too much though, this now only concerns the view, so there's very little technical debt increase if we don't refactor that bit.

CalumTowers:
@NFarrington If you're happy for this to be left as is, I think it gives us a little more flexibility within the view (and saves me a job)

PR #1114, opened 2018-02-04
Logout Button Visibility

atoff:
Unrestricted the logout button from the vatsim-sso guard. Seemed to be causing issues on development builds using the demo sso.

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -97,7 +97,7 @@
                     </li>
                 </ul>
 
-                @if(Auth::check())
+                @if(Auth::guard('vatsim-sso')->check() || Auth::guard('web')->check())
                     {!! Form::open(['route' => 'logout', 'id' => 'logout-form']) !!}
                     <ul class="nav navbar-nav navcustom navbar-right account-dropdown">
                         <li class="dropdown dropdown-large">
@@ -165,7 +165,6 @@
                             </a>
                         </li>
                     </ul>
-                @elseif(Auth::check())
                     {!! Form::open(['route' => 'logout', 'id' => 'logout-form']) !!}
                     <ul class="nav navbar-nav navcustom navbar-right">
                         <li class="dropdown dropdown-large">
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -165,7 +165,7 @@
                             </a>
                         </li>
                     </ul>
-                @elseif(Auth::guard('vatsim-sso')->check())
+                @elseif(Auth::check())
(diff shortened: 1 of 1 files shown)


resources/views/layout.blade.php
```diff
@@ -131,7 +131,7 @@ class="fa fa-sliders"></i> <b class="caret"></b></a>
                                         <li class="divider"></li>
                                         <li>{!! link_to_route("mship.manage.email.add", "Add Email Address") !!}</li>
                                         <li>{!! link_to_route("mship.manage.email.assignments", "Email Assignments") !!}</li>
-                                        @if(Auth::guard('vatsim-sso')->check())
+                                        @if(Auth::check())
```
NFarrington: If this is necessary, then it needs to include both to cover the case where a user has authenticated through `vatsim-sso`, but not `web` (i.e. secondary/full). For example, if a user is half-authenticated, the logout button won't show on the root page `/` in this instance.

`@if(Auth::guard('vatsim-sso')->check() || Auth::guard('web')->check())`

NFarrington:
This is likely an issue with resuming expired (but 'remembered') sessions when you also don't have a secondary password expiry (and therefore are logged in straight away without visiting a `vatsim-sso` guarded page).

resources/views/layout.blade.php
atoff: But is there ever a situation where we don't want to allow an authenticated user to logout? It seems a bit strange restricting logout when a user is logged in! (Unless I am understanding this wrong...)

NFarrington: The purpose of the check is to not display the button to unauthenticated users - so all authenticated or semi-authenticated users should have the button, so that they can log out and wipe their entire session.

atoff: Right, so using check without a guard should be fine, because it only allows (semi) authenticated users to log out? Surely `Auth::check()` accomplishes the same as about, especially if we introduce another guard or something

NFarrington: If you take the primary login page as an example (`/`), this is authenticated through the web guard. If a user is authenticated via `vatsim-sso` but not yet `web`, the logout button will not show for them on this page.

Therefore, the check needs to accommodate the fact that the user may be on a page which uses the `web` guard, but is only logged in via `vatsim-sso`.

Similarly, if a user is logged in only to `web` but not `vatsim-sso` (which will likely occur when a "remembered" session is created), and they visit a page which defaults to the `vatsim-sso` guard, they would still need a log out button.

The current implementation was built assuming that all users who are authenticated as `web` are also authenticated as `vatsim-sso`, which isn't the case for remembered sessions until the user is redirected to a `vatsim-sso` authenticated page, at which point it then remembers that guard, too.

PR #1123, opened 2018-02-11
CORE-33 Documentation Updates

daviesnathan:
This PR contains a number of docs updates in relation to the upcoming migration from GitHub Issues to JIRA.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -165,15 +165,12 @@ If you're comfortable with contributing to Open Source projects on GitHub **plea
 
 It is expected that you will follow the GitFlow Workflow for managing the repository, but here's some important points:
 
-* A new feature should branch `development` into a `feature/<name>` branch
+* A new feature should branch `development` into a `feature/<JIRA-ISSUE-ID>-<name>` branch
 * On completion it should be merged **not rebased** with development
 * You may see us creating `release/<version>` branches - this is where final testing will occur.
 * Where an issue is **assigned** to somebody it means they're working on it.  Speak to them before trying to contribute.
 * If you want to work on an issue. Comment on the issue asking to work on it and a member of the team will add you to the project as a Contributor. This will then allow you to work on the ticket and update it as you go.
 * Any code change **must** have an associated issue detailing (in full) why the change is being made
-* Commit messages **must** reference the ticket and we **recommend** using the smart commit commands available to update your JIRA issue as described [here](https://confluence.atlassian.com/bitbucket/processing-jira-software-issues-with-smart-commit-messages-298979931.html)
- * e.g. To mark the issue for testing and to add a comment `CORE-33 #testing - #comment Added the user's profile picture to their profile.` This will update CORE-33 in JIRA with the status Testing and add the string as a comment to the issue.
- * e.g. To resolve the isue and to add a comment `CORE-33 #done - #comment Added the user's profile picture to their profile.`
 * Commits should ideally be "atomic" in nature
  * A good article explaining atomic commits and their benefits can be viewed [here](https://www.freshconsulting.com/atomic-commits/)
  * Atomic commits allow project maintainers (and you!) to roll back small parts of changes made without having widespread knock-on effects, among other benefits
@@ -195,7 +192,7 @@ When writing your modifications for a single issue, or an entire feature, it is 
 
 There may be times that a test is already written - in these circumstances it is acceptable to contribute without adding an additional test.
 
-## Merge Requests
+## Pull Requests
 
 We welcome merge requests with fixes and improvements to the project.  The features we really would like public support on are marked with ["up-for-grabs"](https://vatsimuk.atlassian.net/browse/CORE-26?jql=project%20%3D%20CORE%20AND%20labels%20%3D%20up-for-grabs%20order%20by%20lastViewed%20DESC) but other improvements are also welcome - please ensure you read over the merge work-flow below.
 
@@ -206,17 +203,16 @@ If you wish to add a new feature or you spot a bug that you wish to fix, **pleas
 The work-flow for submitting a new merge request is designed to be simple, but also ensure consistency from **all** contributors:
 
 * Fork the project into your personal space on GitHub.com
-* Create a new branch (with the name `<issue_number>-<name>`, replacing issue_number with the issue number you're resolving)
- * The exception to this is where an entire feature is being tackled, spanning multiple issues.  In this case you can create a `feature/<name>` branch.
+* Create a new branch. You should include the JIRA issue ID in your branch name (e.g. `CORE-33-<branch name>`).
+ * The exception to this is where an entire feature is being tackled, spanning multiple issues.  In this case you can create a feature branch which links up to the feature's associated JIRA 'Epic' Issue ID (e.g. `feature/CORE-33-<branch name>`)
 * Commit your changes
- * When writing commit messages, consider changing the state your issues via the commit message (by including the issue ID and the state name in your message `CORE-33 #testing #comment Fixed some stuff.`)
 * **Add any important steps that must be followed on deployment to the README.md file**
 * Push the commit(s) to your fork
 * Submit a Pull Request (PR) to **our** development branch
-* The PR title should describe the change that has been made and contain the issue ID. (`ISSUE_ID - SUMMARY`)
+* The PR title should describe the change that has been made.
+  * If you have **not** named your branch in way that links to the task's associated JIRA Issue, you **must** include the JIRA Issue ID in the PR's title. (e.g. `CORE-33 Documentation Updates`). *If your branch is linked to a JIRA issue, your associated Pull Request will be automatically linked*.
 * The PR description should confirm what changes have been made, how you know they're correct (with references)
- * It is expected that all submissions have associated test cases.
-* Ensure you link any relevant issues in the Pull Request by including the Issue ID in the title. This will automatically link the issue to the Pull Request.
+* It is expected that all submissions have associated test cases.
 * Be prepared to answer any questions about your PR when it is reviewed for acceptance.
 
 **If you are actively working on a large change** consider creating the PR early but prefixing it with [WIP] as this will prevent it from being accepted *but* let other people know you're working on that issue.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -151,15 +151,15 @@ If you would like to work on something in the admin panel (which includes some p
...
+* Navigate to `vukcore.localhost/adm/dashboard`. You should now be able to log into the admin panel.
 
 #### Relax
 
 After all that setup, relax for 5 minutes!  If you've had any problems, come and find someone in the Slack team.
 
 ## Contributing to the code
 
-If you're just getting started with GitHub (and project contributions) then we suggest you take a look at issues marked with the "up-for-grabs" label.  These issues will be of reasonable size and challenge, for anyone to start contributing to the project.  [This was inspired by an article by Kent C. Dodds](https://medium.com/@kentcdodds/first-timers-only-78281ea47455#.wior7p101).
+If you're just getting started with GitHub (and project contributions) then we suggest you take a look at issues [marked with the "up-for-grabs" label](https://vatsimuk.atlassian.net/browse/CORE-26?jql=project%20%3D%20CORE%20AND%20labels%20%3D%20up-for-grabs%20order%20by%20lastViewed%20DESC).  These issues will be of reasonable size and challenge, for anyone to start contributing to the project.  [This was inspired by an article by Kent C. Dodds](https://medium.com/@kentcdodds/first-timers-only-78281ea47455#.wior7p101).
```
CalumTowers: Will we be masking the domain with a CNAME? If so, can we update the link to reflect that please?

CONTRIBUTING.md
```diff
@@ -206,28 +197,29 @@ There may be times that a test is already written - in these circumstances it is
 
 ## Merge Requests
 
-We welcome merge requests with fixes and improvements to the project.  The features we really would like public support on are marked with "up-for-grabs" but other improvements are also welcome - please ensure you read over the merge work-flow below.
+We welcome merge requests with fixes and improvements to the project.  The features we really would like public support on are marked with ["up-for-grabs"](https://vatsimuk.atlassian.net/browse/CORE-26?jql=project%20%3D%20CORE%20AND%20labels%20%3D%20up-for-grabs%20order%20by%20lastViewed%20DESC) but other improvements are also welcome - please ensure you read over the merge work-flow below.
```
CalumTowers: Same here

CONTRIBUTING.md
```diff
@@ -206,28 +197,29 @@ There may be times that a test is already written - in these circumstances it is
...
+
+If you wish to add a new feature or you spot a bug that you wish to fix, **please open an issue for it first** in the [CORE project](https://vatsimuk.atlassian.net/projects/CORE/issues).
```
CalumTowers: Same here

CONTRIBUTING.md
daviesnathan: Custom domains and CNAMEs are not currently supported by Atlassian's Cloud Offerings. I have spoken to @NFarrington about implementing some basic sub-domain rewrite rules to redirect users to the site - not sure how far that is out so I would say it is probably best to leave them as is for now and when we implement the rewrites I can quickly update then.

What do you think? @CalumTowers

CalumTowers: @daviesnathan Disappointing to learn that enterprise grade software won't allow white labelling... We should keep an eye on https://jira.atlassian.com/browse/CLOUD-6999.

daviesnathan: Completely agree, I have voted on it to try and bump it up and I am actively watching it. Apparently they have it in a Private Beta at the moment so shouldn't be too long, the rewrites along with parameter passing should give a similar experience in the meantime.

CONTRIBUTING.md
```diff
@@ -169,13 +169,11 @@ It is expected that you will follow the GitFlow Workflow for managing the reposi
...
- * We would advise against working on issues without an Acknowledged label.
- * If you want to start work on an issue, comment on the issue asking to work on it and it'll be assigned to you by a developer.
+* If you want to work on an issue. Comment on the issue asking to work on it and a member of the team will add you to the project as a Contributor. This will then allow you to work on the ticket and update it as you go.
 * Any code change **must** have an associated issue detailing (in full) why the change is being made
-* Commit messages **must** make use of the issue management commands
- * When resolving an issue via a commit `Fixes #123 - Added the user's profile picture to their profile.`
- * When addressing an issue via a commit `Restructured the user's profile page to make room for the badges in #456`
+* Commit messages **must** reference the ticket and we **recommend** using the smart commit commands available to update your JIRA issue as described [here](https://confluence.atlassian.com/bitbucket/processing-jira-software-issues-with-smart-commit-messages-298979931.html)
```
NFarrington: I am hesitant to say commit messages **must** reference the ticket, since this will clutter the GitHub commit messages with info that is likely unnecessary. Especially if we have a large feature, we wouldn't want every commit to reference an issue.

CONTRIBUTING.md
```diff
@@ -169,13 +169,11 @@ It is expected that you will follow the GitFlow Workflow for managing the reposi
...
+ * e.g. To mark the issue for testing and to add a comment `CORE-33 #testing - #comment Added the user's profile picture to their profile.` This will update CORE-33 in JIRA with the status Testing and add the string as a comment to the issue.
```
NFarrington: Building on the above, I don't believe we want to add a comment to the ticket for every commit message, or make the commit message the same as the comment. The ticket should already be linked to the branch and pull request, so a comment is actually needed, this probably isn't necessary.

NFarrington: Similarly, to reduce clutter, might it be worth advising people to only use smart commit messages in the main body of a commit message (not the summary)? We ideally want the summary to only include a brief message about what this commit does (and that brevity will be shortened further by the issue tag and smart commit tags).

In particular, https://chris.beams.io/posts/git-commit/ has always been a good reference for deciding how to format commit messages (we could even include it in the contributing guide somewhere).

CONTRIBUTING.md
```diff
@@ -206,28 +197,29 @@ There may be times that a test is already written - in these circumstances it is
...
-* Submit a merge request (MR) to **our** development branch
-* The MR title should describe the change that has been made
-* The MR description should confirm what changes have been made, how you know they're correct (with references)
+* Submit a Pull Request (PR) to **our** development branch
+* The PR title should describe the change that has been made and contain the issue ID. (`ISSUE_ID - SUMMARY`)
+* The PR description should confirm what changes have been made, how you know they're correct (with references)
  * It is expected that all submissions have associated test cases.
-* Ensure you link any relevant issues in the merge request (you can type hash and the issue ID, eg #275).  Comment on those issues linking back to the MR (you can reference MRs using the format !<MR_ID> for example !22).
-* Be prepared to answer any questions about your MR when it is reviewed for acceptance.
+* Ensure you link any relevant issues in the Pull Request by including the Issue ID in the title. This will automatically link the issue to the Pull Request.
```
NFarrington: Does JIRA support linking these in the description, rather than the title?

CONTRIBUTING.md
```diff
@@ -206,28 +197,29 @@ There may be times that a test is already written - in these circumstances it is
...
+
+If you wish to add a new feature or you spot a bug that you wish to fix, **please open an issue for it first** in the [CORE project](https://vatsimuk.atlassian.net/projects/CORE/issues).
 
-If you wish to add a new feature or you spot a bug that you wish to fix, **please open an issue for it first** on the [UK Core Web Services Repository](https://github.com/vatsim-uk/core/issues).
+**Note:** You will need to signup to create issues within the project. See [Issue Tracking](ISSUE_TRACKING.md) for more information.
 
 The work-flow for submitting a new merge request is designed to be simple, but also ensure consistency from **all** contributors:
 
 * Fork the project into your personal space on GitHub.com
 * Create a new branch (with the name `<issue_number>-<name>`, replacing issue_number with the issue number you're resolving)
```
NFarrington: It might be worth including a JIRA-specific example here now, e.g. the branch will be called `core-33` or `core-33-documentation-updates

CONTRIBUTING.md
```diff
@@ -206,28 +197,29 @@ There may be times that a test is already written - in these circumstances it is
...
  * The exception to this is where an entire feature is being tackled, spanning multiple issues.  In this case you can create a `feature/<name>` branch.
```
NFarrington: If working on a feature branch, will these also be linked to an issue if we call them `feature/core-##-my-great-feature`?

CONTRIBUTING.md
daviesnathan: Sure, I had done this elsewhere but missed this part.

CONTRIBUTING.md
daviesnathan: Okay - fair enough. I agree that **must** is a bit too far, I think the option should remain however, for example in the case that someone is working on a feature, they should reference the feature's associated "Epic" issue in JIRA when creating the branch and PR. If individual issues within said epic are addressed in a particular commit, I would like to see them there.

daviesnathan: Thinking on this actually, commit-level smart functions such as issue links and comments do not actually work against our current workflow. As the repo is currently forked and merged back into, it would appear that only the branches and PRs are picked up by JIRA's link to the Core Repository as that is the point that code starts to be moved back into the Core Repo.

CONTRIBUTING.md
daviesnathan: Fair enough, I will remove the reference to Smart Comments. As said above, not sure they work, they haven't for the two commits within this PR.

daviesnathan: As above, no longer relevant - as they appear not to work. (Only found this out as of this PR)

CONTRIBUTING.md
daviesnathan: Unfortunately not. However, upon further reading, if the source branch of the Pull Request is linked to the issue, then the PR will automatically be linked. So I will update accordingly.

CONTRIBUTING.md
daviesnathan: Yes. Although I would reckon that in the case of a feature, the Epic issue should be linked and not the individual issues within it.

NFarrington: Agree completely 👍

PR #1136, opened 2018-02-26
Hotfix: Accommodate slugs in permissions

atoff:
Utilise regex to search for alphanumeric slugs. Will be redone later on during review of permissions logic and utilising policies, but until then this will suffice.

### DIFF
--- a/app/Models/Mship/Role.php
+++ b/app/Models/Mship/Role.php
@@ -138,6 +138,7 @@ class Role extends Model
             // Replace wildcard
             $perm_has = str_replace('*', '[A-Za-z0-9]+', $perm_has);
             $perm_has = '/^'.$perm_has.'$/';
+            
             return (boolean) preg_match($perm_has, $permission);
         })->isEmpty();
     }
--- a/app/Models/Mship/Role.php
+++ b/app/Models/Mship/Role.php
@@ -128,7 +128,21 @@ class Role extends Model
         return !$this->permissions->filter(function ($perm) use ($permission) {
             $stripped = preg_replace("/[^A-Za-z0-9\/]/i", '', $perm->name);
 
-            return strcasecmp($perm->name, $permission) == 0 || strcasecmp($stripped, $permission) == 0 || $perm->name == '*';
+            if(strcasecmp($perm->name, $permission) == 0 || strcasecmp($stripped, $permission) == 0 || $perm->name == '*'){
+              return true;
+            }
+
+            // Secondary fallback - Mainly for non-numeric slugs
+            // Add slashes
+            $perm_has = str_replace('/', '\/', $perm->name);
+            // Replace wildcard
+            $perm_has = str_replace('*', '[A-z0-9]+', $perm_has);
+            $perm_has = "/" . $perm_has . "$/";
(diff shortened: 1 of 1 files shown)


app/Models/Mship/Role.php
```diff
@@ -128,7 +128,21 @@ public function hasPermission($permission)
...
-            return strcasecmp($perm->name, $permission) == 0 || strcasecmp($stripped, $permission) == 0 || $perm->name == '*';
+            if (strcasecmp($perm->name, $permission) == 0 || strcasecmp($stripped, $permission) == 0 || $perm->name == '*') {
+                return true;
+            }
+
+            // Secondary fallback - Mainly for non-numeric slugs
+            // Add slashes
+            $perm_has = str_replace('/', '\/', $perm->name);
+            // Replace wildcard
+            $perm_has = str_replace('*', '[A-z0-9]+', $perm_has);
```
NFarrington: Should be `[A-Za-z0-9]+` - `A-z` will include additional characters which shouldn't be expected.

app/Models/Mship/Role.php
```diff
@@ -128,7 +128,21 @@ public function hasPermission($permission)
...
+            $perm_has = '/'.$perm_has.'$/';
```
NFarrington: Should be `'/^'.$perm_has.'$/'` to prevent partial matches (unlikely but possible in the future)

app/Models/Mship/Role.php
```diff
@@ -128,7 +128,21 @@ public function hasPermission($permission)
...
+            if (preg_match($perm_has, $permission)) {
+                return true;
```
NFarrington: Could be shortened to `return (boolean) preg_match($perm_has, $permission);` to squash L141-145, if you wanted - up to you though.

app/Models/Mship/Role.php
```diff
@@ -128,7 +128,21 @@ public function hasPermission($permission)
...
             $stripped = preg_replace("/[^A-Za-z0-9\/]/i", '', $perm->name);
 
-            return strcasecmp($perm->name, $permission) == 0 || strcasecmp($stripped, $permission) == 0 || $perm->name == '*';
+            if (strcasecmp($perm->name, $permission) == 0 || strcasecmp($stripped, $permission) == 0 || $perm->name == '*') {
+                return true;
+            }
+
+            // Secondary fallback - Mainly for non-numeric slugs
+            // Add slashes
+            $perm_has = str_replace('/', '\/', $perm->name);
```
NFarrington: Can this be done with `preg_quote()`, to escape all regex characters? https://secure.php.net/manual/en/function.preg-quote.php

atoff: According to the docs `Note that / is not a special regular expression character.`. Did test it as well, but it didn't modify the string

NFarrington: The `$delimiter` parameter can be used for that purpose
>If the optional delimiter is specified, it will also be escaped. This is useful for escaping the delimiter that is required by the PCRE functions. The / is the most commonly used delimiter.

atoff: Must of missed that, updated

PR #1146, opened 2018-03-02
Various Feedback Fixes & Improvements

### LINKED ISSUE #1122: Radio-Type Questions can be created with no values
On creating a new feedback form, radio type questions can have their values left empty, which causes an error when trying to display them.

https://github.com/VATSIM-UK/core/blob/production/app/Http/Controllers/Adm/Mship/Feedback.php

Put in a check to see if the field requires values, and check we do actually have some.

### LINKED ISSUE #1126: Feedback Form Question's Slug length too small
## Error in VATSIM-UK/Core (Laravel)

**Illuminate\Database\QueryException** in **POST /adm/mship/feedback/configure/atcmentor**
SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'slug' at row 1 (SQL: update `mship_feedback_questions` set `slug` = additionalcomments110, `required` = 0, `updated_at` = 2018-02-11 17:37:56 where `id` = 45)

[View on Bugsnag](https://app.bugsnag.com/vatsim-uk/vatsim-uk-slash-core-laravel/errors/5a807e9523abaf00187a9952?event_id=5a807f75fcc9b20018e8df88&i=gh&m=ci)

## Stacktrace

    app/Http/Controllers/Adm/Mship/Feedback.php:106 - App\Http\Controllers\Adm\Mship\Feedback::configureForm
    app/Http/Controllers/Adm/Mship/Feedback.php:60 - App\Http\Controllers\Adm\Mship\Feedback::postConfigure
    app/Http/Controllers/BaseController.php:38 - App\Http\Controllers\BaseController::App\Http\Controllers\{closure}
    app/Http/Middleware/RedirectToIntended.php:23 - App\Http\Middleware\RedirectToIntended::handle
    app/Traits/Middleware/RedirectsOnFailure.php:48 - App\Http\Middleware\MustHaveCommunityGroup::handle
    app/Traits/Middleware/RedirectsOnFailure.php:48 - App\Http\Middleware\UserMustReadNotifications::handle
    app/Http/Middleware/DenyIfBanned.php:27 - App\Http\Middleware\DenyIfBanned::handle
    app/Traits/Middleware/RedirectsOnFailure.php:48 - App\Http\Middleware\MandatoryPasswords::handle
    app/Http/Middleware/RecordLoginInfo.php:39 - App\Http\Middleware\RecordLoginInfo::handle
    app/Http/Middleware/TrackInactivity.php:43 - App\Http\Middleware\TrackInactivity::handle

[View full stacktrace](https://app.bugsnag.com/vatsim-uk/vatsim-uk-slash-core-laravel/errors/5a807e9523abaf00187a9952?event_id=5a807f75fcc9b20018e8df88&i=gh&m=ci)

*Created by Alex Toff via Bugsnag*

### LINKED ISSUE #1145: Feedback forms with no contact cause error on summary email
When a feedback form doesn't have a contact assigned, the summary email job errors.

Either force feedback forms to have a contact assigned, or alter logic in job.

atoff:
1) Removal of sequence number from form slug (No use now that sequence is itself a column)
2) Introduce validation for question types that require values to be set. Fixes #1122 
3) Increase form slug to 255 chars. Fixes #1126 
4) Restyling of feedback configuration page, making it understandable and less cluttered.
image
5) Move the various feedback links from the sidebar into a standalone page. De-clutters the sidebar. Also, adds a little New icon when un-actioned feedback exists for forms that the user can view.
image
6) Check that contact exists for feedback summary job. If it doesn't exist, no email is sent. Fixes #1145

### DIFF
--- a/resources/views/adm/mship/feedback/settings.blade.php
+++ b/resources/views/adm/mship/feedback/settings.blade.php
@@ -109,10 +109,29 @@
                                 <div class="box-header">
                                     <h4 class="box-title" style="font-size:1.5em">
                                         Form Questions
-                                    </h4>
+                                    </h4></br>
+                                    <small><b>Note:</b> You do NOT need to add a 'userlookup' question if the form is targeted. It is added automatically</small>
                                 </div>
                                 <div class="box-body feedback-form-config">
                                     <div class="row">
+                                        @if ($form->targeted)
+                                          <div class="col-md-12 permanent question-item">
+                                            <div class="box">
+                                              <div class="box-header">
+                                                <div class="row">
+                                                  <div class="col-md-8">
+                                                    <div class="input-group">
+                                                      <span class="input-group-addon" id="question-name-addon"><b>Question</b></span>
+                                                      {{ Form::text('', 'CID of the member you are leaving feedback for.', ['aria-describedby' => 'question-name-addon', 'size' => 50, 'disabled']) }}
+                                                      <span class="input-group-addon">(Permanent)</span>
+                                                    </div>
+
+                                                  </div>
+                                                </div>
+                                              </div>
+                                            </div>
+                                          </div>
+                                        @endif
                                         <ol class='simple_connected_list' id="feedback-form-questions">
                                             @if (old('old_data') != null)
                                                 {!! old('old_data') !!}
--- a/resources/views/adm/mship/feedback/settings.blade.php
+++ b/resources/views/adm/mship/feedback/settings.blade.php
@@ -1,27 +1,5 @@
 @extends('adm.layout')
 
-@section('styles')
-<style>
-    /* JQuery Sortable Styling */
-    body.dragging, body.dragging * {
-        cursor: move !important;
-    }
-
-    .dragged {
-        position: absolute;
-        opacity: 0.5;
-        z-index: 2000;
-    }
-    ol.simple_connected_list{
-        display: inline;
-    }
-
-    ol.simple_connected_list li {
-        list-style: none;
-    }
-</style>
-@endsection
-
 @section('scripts')
     <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-sortable/0.9.13/jquery-sortable-min.js"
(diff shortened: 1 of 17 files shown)


NFarrington:
When clicking the Feedback button in the sidebar and visiting Feedback pages, the Membership menu opens on each page load. This exists in production, however with no menu for Feedback, this would need fixing.

We should keep the dropdown menu as-is, with the settings links moved to a sub-menu (similar to the smartCARS configuration menu). The `NEW` label isn't useful for people with access to many forms, and ideally we'll instead have the number of unactioned feedback for each form at the side of each link in the sidebar (similar to VT applications).

We also need to retain accessibility to the full list of feedback across all forms.

For editing forms, if I add two new elements to the form, and then click the settings button on the bottom one, it will always expand the top one instead.

atoff:
In terms of keeping the drop-down menu I think, especially as we start to add more forms, using a clean, table format will be much cleaner in the long run, as the sidebar submenu would end up getting too long, it already looks cluttered at the moment imo, even if we didn't have the settings links. What I could do, however, is if the user only has access X (say 3/4) forms, then we can display it in a submenu, else in the table format?. Agree on the other points, and working on fixes

NFarrington:
We wouldn't want it to appear different for different users, unless the user had the control to configure it. The implementations I have in mind for fixing the problem you're suggesting will ideally be dealt with after our refactoring, so can we maybe revisit it further down the line? It is something that will need dealing with, but I don't think the list is currently big enough that it needs implementing within the next few months. Are you happy to hold off for now, and we'll deal with it with the new processes and such we're implementing in JIRA?

atoff:
Should all be addressed now @NFarrington ?

resources/views/adm/mship/feedback/settings.blade.php
```diff
@@ -105,6 +81,13 @@
             $("#feedback-form-questions").on("click", ".question-delete-button", function () {
                 $(this).closest('.question-item').remove();
             });
+
+            // Question accordion control
+            $("#feedback-form-questions").on( 'click', '.question-settings-control', function () {
```
NFarrington: This is only present on `settings.blade.php` - needs copying over for new forms.

NFarrington:
Merged in 413d4503472c7c327401606c5f035b491b90de82

PR #1152, opened 2018-03-07
Hotfix 3.6.5

NFarrington:
Resolves:

CORE-38 - Remove incorrect TeamSpeak foreign key
CORE-51 - Fix issue searching PIREPs for those with a specific departure/arrival ID
CORE-52 - Fix smartCARS POSREPs using commas instead of periods in their lat/lons.
CORE-53 - Fix displaying of VT applications where a user's previous rating is unknown.
CORE-56 - Lapse VT applications with expired referees.


NFarrington:
Ready to deploy on Thursday 8th March.

PR #1159, opened 2018-04-20
CORE-40 Refine FTE Fail Reasons

CalumTowers:
Resolves CORE-40


CalumTowers:
@NFarrington Fixed. Looks like I forgot how foreign keys worked... 😂

CalumTowers:
@NFarrington Yes, it works. Its exactly the same logic just split into three functions rather than one (`matchesCriteria`)

CalumTowers:
That’s what we want? If it fails on any posrep in any criteria the entire flight fails?

CalumTowers:
It can’t be failing even after the first criterion because I’ve had tests run through a live flight with real data and it passes

CalumTowers:
The code actually checks for the posrep matching any of the criteria (in no particular order). This was amended prior to the initial release

NFarrington:
It *used* to check if a posrep matched any criteria - it's now checking if a posrep matches *all* criteria. Please follow the logic through to see what I'm saying.

The code says: For every posrep in the flight, check it against every criteria in the flight. If this posrep I am currently checking fails against the criteria, stop, otherwise, check it against the next criteria. If it fails against this criteria, stop, otherwise check it against the next criteria, and so on.

What it SHOULD say is: For every posrep in the flight, check it against every criteria in the flight. If this posrep I am currently checking **passes** against the criteria, continue to the next posrep. If we've checked all criteria and it hasn't passed any of them, then stop and fail.

If you still insist the code works, please send me some test data that you believe shows it to be working.

CalumTowers:
@NFarrington Good shout, done

PR #1161, opened 2018-04-22
CORE-57 - Check for $account object

CalumTowers:
Resolves CORE-57

### DIFF
--- a/app/Http/Controllers/Smartcars/Api/Authentication.php
+++ b/app/Http/Controllers/Smartcars/Api/Authentication.php
@@ -37,7 +37,7 @@ class Authentication extends AdmController
             return 'AUTH_FAILED';
         }
 
-        if ($account->is_banned) {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Smartcars/Api/Authentication.php
```diff
@@ -37,7 +37,7 @@ public function postManual()
             return 'AUTH_FAILED';
         }
 
-        if ($account->is_banned) {
+        if ($account && $account->is_banned) {
```
NFarrington: This is simply cascading the error further down the code, so it will now occur on line 44 instead, when `->verifyPassword()` is called on the same object.

We should simply return AUTH_FAILED when the account is first retrieved if it's null - then we can remove any further `$account` checks in the function, which will mean the same error isn't made in future when someone writes another line which depends upon the `$account` object.

PR #1171, opened 2018-06-24
CORE-55 - Referee notification

atoff:
Fixed logic that deletes references & notifies referees. Once the application is withdrawn or expired, the reference and its access tokens are deleted, and the referee is emailed.

Closed CORE-55

### DIFF
--- a/app/Listeners/VisitTransfer/NotifyRefereeOfReferenceCancellation.php
+++ b/app/Listeners/VisitTransfer/NotifyRefereeOfReferenceCancellation.php
@@ -3,10 +3,10 @@
 namespace App\Listeners\VisitTransfer;
 
 use App\Events\VisitTransfer\ReferenceCancelled;
-use App\Notifications\ApplicationReferenceCancelled;
+use App\Notifications\ApplicationReferenceNoLongerNeeded;
 use Illuminate\Contracts\Queue\ShouldQueue;
 
-class NotifyApplicantOfReferenceCancellation implements ShouldQueue
+class NotifyRefereeOfReferenceCancellation implements ShouldQueue
 {
     public function __construct()
     {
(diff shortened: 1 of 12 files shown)


atoff:
With the token, the options are either delete it and deal with the 404 (a custom message like that it would require a large refactoring of code - technically a 404 is correct) or keep the token and display a reference cancelled once the referee uses the token. Personally, I think that receiving an email is enough, and we should get rid of the unnecessary token entries in the database once cancelled. 

I will look into the cancelling the reference instead though

CalumTowers:
@atoff The token itself could be deleted, but the reference itself be marked as status 100 and leave deleted_at as null

I've not reviewed the token handling so not sure how that would play out right now

atoff:
@CalumTowers  this should be what you wanted? Due to the way the whole cancellation logic was written, it was assumed that it was only for when the referee said they didn't know the application. I have re-jigged it a bit so it is more generic now to fit what we need.

app/Listeners/VisitTransfer/NotifyRefereeOfReferenceCancellation.php
```diff
@@ -15,7 +15,6 @@ public function __construct()
 
     public function handle(ReferenceCancelled $event)
     {
-        $reference = $event->reference;
-        $reference->application->account->notify(new ApplicationReferenceCancelled($reference));
+        $event->reference->notify(new ApplicationReferenceNoLongerNeeded($event->reference));
```
CalumTowers: We should ensure that this event only fires to relevant referees (using the static method REFERENCE_IS_PENDING) to avoid sending users unnecessary notifications

atoff: This already happens elsewhere, see https://github.com/VATSIM-UK/core/pull/1171/files#diff-fda47f2e5dc0e1b3b11e181ac8d7a0fdR279

atoff:
@CalumTowers I have fixed the last few things. I could not replicate a cancelled referee receiving a notification when the application is rejected, only references that were still pending are receiving them.

CalumTowers:
@atoff Disagree, `$reference->cancel();` in the foreach is not excluding certain stati unless I'm missing something? To replicate, open application, login as referee and say do not know applicant, go to adm, reject application, referee gets notification.

Could we make the note simpler, see below
note

CalumTowers:
Thanks @atoff 👍

PR #1177, opened 2018-06-28
Add 5.6 Logging

CalumTowers:


config/logging.php
```diff
@@ -0,0 +1,70 @@
...
+        ],
+        'daily' => [
+            'driver' => 'daily',
+            'path' => storage_path('logs/laravel.log'),
+            'level' => 'debug',
+            'days' => 7,
+        ],
+        'slack' => [
+            'driver' => 'slack',
+            'url' => env('LOG_SLACK_WEBHOOK_URL'),
```
atoff: Seeing as this doesn't have a default, should we consider adding it to the `.env.example`? Alternatively, we could remove it/comment it out until we need it? Up to you but happy to merge without.

CalumTowers: @atoff Good shout. Added in 7bac7f7

PR #1179, opened 2018-06-30
[WIP] Quarterly Stats

CalumTowers:


app/Http/Controllers/Adm/Operations/QuarterlyStats.php
```diff
@@ -0,0 +1,99 @@
...
+        return $query;
+    }
+
+    private function completedTransfersExObs($startDate, $endDate) {
+        $query = DB::table('mship_account_state')
+            ->where('state_id', '=', 3)
+            ->whereNull('end_at')
+            ->whereBetween('start_at', [$startDate, $endDate])
+            ->whereIn('account_id', function($startDate, $endDate, $states) {
+                $states->select('account_id') // too few variables exception thrown
```
CalumTowers: @atoff Not sure if I'm being a little thick but an exception is being thrown for number of variables here...

CalumTowers:
@atoff I've literally just grabbed the raw SQL and put it into here. Not really using any existing relationships at all.

Fancy helping optimise and tidy?

app/Http/Controllers/Adm/Operations/QuarterlyStats.php
AxonC: Do you need a double equality for state ID? (==)

I’d imagine your $states var could be null at some point? Could you dd() each part perhaps?

atoff:
@CalumTowers take a look at the new statements to make sure they do what you want them to do, then it is up to you for front end ;)

CalumTowers:
@AxonC Can you sanity check this please? There is a lack of testing at the moment...

PR #1185, opened 2018-07-05
Lock TeamSpeak Package

CalumTowers:
Resolves issue on v3.8.0 where TeamSpeak failing connection

Further investigation required to updated to latest package


CalumTowers:
@AxonC @atoff Can either of you just sanity check this please, locking to known working version

PR #1190, opened 2018-07-12
[WIP] Visiting Controllers Monitoring Tool

AxonC:
Resolves CORE-45


AxonC:
Tests perhaps pending - trying to keep the work below the estimate.

app/Http/Controllers/VisitTransfer/Admin/VisitorStatsController.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        parent::__construct();
+
+        $this->accounts = $accounts;
+    }
+
+    public function index()
+    {
+        $inputStartDate = Input::get('startDate');
+        $inputEndDate = Input::get('endDate');
```
CalumTowers: Can we take advantage of Carbon/Request here so that a date can be inputted but it converts it correctly for the query?
Like the below?

https://github.com/VATSIM-UK/core/blob/93c0085512563fbda82fb1139553c383c4280fd7/app/Http/Controllers/Adm/Mship/Feedback.php#L267-L274

AxonC: We could if it was a post request - the Date submits via a get request. I'm happy to do a new carbon object rather than parsing it though if that is what you're getting at?

CalumTowers: I'd rather we just posted, there isn't any real benefit in doing it via GET as I doubt people will bookmark the URL with the query string.

Perhaps we could simplify this a little then by forcing the user to enter a date before an output is displayed? That way, we can just grab the request, perform the query and return rather than defaulting the `$from_date` and `$to_date`

CalumTowers:
@AxonC I've made some minor amends.

However, some of this code is very heavy on DB usage. My dev env has 1 user with 500 rows of networkdata_atc and the code executed over 1000 queries on this page.

Can we look to optimise please?

PR #1191, opened 2018-07-16
Slack Error Handling

atoff:
Closes CORE-90

### DIFF
--- a/app/Console/Commands/Command.php
+++ b/app/Console/Commands/Command.php
@@ -3,10 +3,10 @@
 namespace App\Console\Commands;
 
 use App;
-use Slack;
 use Bugsnag;
 use GuzzleHttp\Exception\BadResponseException;
 use Illuminate\Console\Command as BaseCommand;
+use Slack;
 use Symfony\Component\Console\Output\OutputInterface;
 
 abstract class Command extends BaseCommand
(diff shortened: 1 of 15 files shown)


atoff:
@CalumTowers can you check that this is the kind of thing you are after?

CalumTowers:
@atoff Yeah this is where we need to go. I want to try and eliminate the "its Slack's fault" from our BugSnag to...

a) tidy it up
b) stop using our quota

I mean, we do need to optimise our calls to Slack's API to prevent us being blocked/timed out but if its a temporary error with no real impact to the user, I don't want BugSnag (and my phone...) going crazy

app/Console/Commands/Command.php
```diff
@@ -206,4 +216,35 @@ protected function getAuthorLink()
             return 'https://gitlab.com/vatsim-uk/core/blob/development'.$directory;
         }
     }
+
+    private function handleSlackException($e)
+    {
+        $errorClass = get_class($e);
+        $errorCode = $e->getCode();
+        if ($errorClass == ClientException::class) {
```
CalumTowers: I may be being a little thick here but what is the purpose of checking the error class? Can we not just grab the code, if 408 or 504, break else BugSnag?

atoff: Good point. I will make it catch the `GuzzleHttp\Exception\BadResponseException` which will catch both.

PR #1194, opened 2018-07-19
Prevent self feedback

CalumTowers:
Resolves CORE-107


CalumTowers:
@atoff Good point... I've addressed that in 9ee9fd9.

Its messy but we could probably do with revisiting that controller to refactor. Slightly outside of the scope of this PR.

PR #1202, opened 2018-07-22
[WIP] Students in Waiting List

AxonC:
Resolves CORE-97

### DIFF
--- a/app/Models/Training/WaitingList.php
+++ b/app/Models/Training/WaitingList.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace App\Models\Training;
+
+use App\Models\Mship\Account;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
+
+class WaitingList extends Model
+{
+    use SoftDeletes;
+
+    public $table = 'training_waiting_list';
+
+    protected $dates = ['deleted_at'];
+
+    protected $casts = ['active' => 'boolean'];
+
+    public function getRouteKeyName()
+    {
+        return 'slug';
+    }
+
+    /**
+     * Many WaitingLists can have many Accounts (pivot).
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsToMany
+     */
+    public function students()
+    {
+        return $this->belongsToMany(Account::class, 'training_waiting_list_account',
+            'list_id')->using(WaitingListAccount::class);
+    }
+
+    /**
+     * Add an Account to a waiting list.
+     *
+     * @param Account $account
+     */
+    public function addToWaitingList(Account $account)
+    {
+        return $this->students()->attach($account);
+    }
+
+    /**
+     * Remove an Account from a waiting list.
+     *
+     * @param Account $account
+     * @return int
+     */
+    public function removeFromWaitingList(Account $account)
+    {
+        return $this->students()->detach($account);
+    }
+
+    /**
+     * Retrieve all active waiting lists.
+     *
+     * @param \Illuminate\Database\Eloquent\Builder $query
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopeActive($query)
+    {
+        return $query->where('active', true);
+    }
+}
--- a/tests/Unit/Training/WaitingListTest.php
+++ b/tests/Unit/Training/WaitingListTest.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Tests\Unit\Training;
+
+use App\Models\Mship\Account;
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+use Tests\TestCase;
+
+class WaitingListTest extends TestCase
+{
+    use DatabaseTransactions;
+
+    private $waitingList;
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->waitingList = factory(\App\Models\Training\WaitingList::class)->create();
+    }
+
+    /** @test * */
+    public function itHasASlugRouteKey()
+    {
+        $this->assertEquals('slug', $this->waitingList->getRouteKeyName());
+    }
+
+    /** @test * */
+    public function itHasAName()
+    {
+        $this->assertNotNull($this->waitingList->name);
+        $this->assertNotNull($this->waitingList->slug);
+    }
+
+    /** @test * */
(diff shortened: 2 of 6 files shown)


AxonC:
This PR will be focusing on the setup of the backend structure for this epic.

AxonC:
This PR covers the extent of the quoted issue, @CalumTowers, and if anything goes beyond. I'd like to merge to the epic branch so I can start work on CORE-108

routes/web-admin.php
```diff
@@ -223,3 +223,14 @@
         'uses' => 'VisitorStatsController@index',
     ]);
 });
+
+Route::group([
```
CalumTowers: Can you format this using the `->uses()`  and `->name()` helpers please? This will allow easier merging with #1219.

tests/Unit/Training/WaitingListTest.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /** @test * */
+    public function itHasAName()
+    {
+        $this->assertNotNull($this->waitingList->name);
+        $this->assertNotNull($this->waitingList->slug);
+    }
+
+    /** @test * */
+    public function itCanHaveStudents()
```
CalumTowers: You hinted at test name standardisation in another PR; I totally agree. Can we decide on what we will call people and then we will look to update existing tests in another PR? Students/Members/Users

app/Models/Training/WaitingList.php
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Demote an Accounts' position within a WaitingList by the number passed.
+     *
+     * @param Account $account
+     * @param int $position
+     */
+    public function demote(Account $account, $position = 1)
+    {
+        if (!$this->accounts->contains($account)) {
```
CalumTowers: Would this be cleaner by being nested within a try catch? Perhaps using the `findOrFail` method on the account model?

_(comment applicable for `promote` too)_

app/Models/Training/WaitingList.php
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    /**
+     * Demote an Accounts' position within a WaitingList by the number passed.
+     *
+     * @param Account $account
+     * @param int $position
+     */
+    public function demote(Account $account, $position = 1)
```
CalumTowers: Can we fire an event from this function to log it? Happy for you to just fire an event for now and then implement in another branch?

I'm not sure if we already have something defined already in our code base.

_(comment applicable for `promote` too)_

app/Models/Training/WaitingList.php
```diff
@@ -0,0 +1,170 @@
...
+        $this->accounts()->attach($account, ['position' => $this->nextPosition(), 'added_by' => $staffAccount->id]);
+    }
+
+    /**
+     * Remove an Account from a waiting list.
+     *
+     * @param Account $account
+     * @return void
+     */
+    public function removeFromWaitingList(Account $account)
```
CalumTowers: Can we log this?

app/Models/Training/WaitingList.php
```diff
@@ -0,0 +1,170 @@
...
+            'list_id')->using(WaitingListAccount::class)->withPivot(['position'])->withTimestamps();
+    }
+
+    /**
+     * Add an Account to a waiting list.
+     *
+     * @param Account $account | Collection
+     * @param Account $staffAccount
+     */
+    public function addToWaitingList($account, Account $staffAccount)
```
CalumTowers: Can we log this?

resources/views/adm/training/index.blade.php
```diff
@@ -0,0 +1,36 @@
...
+    <section class="content">
+        <div class="row">
+            <div class="col-xs-12">
+                <div class="box box-warning">
+                    <div class="box-header">
+                        <div class="box-title">Waiting List Management</div>
+                    </div>
+                    <div class="box-body">
+                        <div class="row">
+                            <div class="col-xs-12">Select a Waiting List to Manage {{ $lists }}</div>
```
CalumTowers: Can we remove the `{{ $lists }}` before merge?

resources/views/adm/training/manage.blade.php
```diff
@@ -0,0 +1,88 @@
...
+                            </tbody>
+                        </table>
+                    </div>
+                </div>
+            </div>
+
+            <div class="modal fade" id="modalWaitingListAdd" tabindex="-1" role="dialog">
+                <div class="modal-dialog">
+                    {!! Form::open(['route' => ['training.waitingList.store', $waitingList], 'method' => 'post']) !!}
+                    <div class="modal-content">
```
CalumTowers: Can we use the same modal that is used for feedback? This will use a nice JS popup that returns a list of people (and allows searching by name)

Ref: `route('mship.feedback.usersearch')`

tests/Unit/Training/WaitingListTest.php
AxonC: Generally, I think when testing we should use the name on the user story so that we can assert in our own minds that we are actually meeting the requirements. Thoughts on that?

CalumTowers: Yep, that makes sense. We should ensure it is consistent throughout the test/feature in that case

app/Models/Training/WaitingList.php
AxonC: My only concern with this is what behaviour do we want to happen? We obviously don't want to trigger redirects within the Model and this method is only called in a controller which catches this exception if that makes sense?

CalumTowers: Should we not just guard against accounts not being found within the controller before it even hits the model? I'm not _too_ worried about this but we can defo remove the `else` and just have the `if`

AxonC: It can always be refactored when we get to the next branch but I'll remove the `if` because that **is** ugly.

resources/views/adm/training/manage.blade.php
AxonC: Per Slack discussion - that modal is not available in adm due to dependencies within layout - suggest implementing at later date @CalumTowers

CalumTowers:
@AxonC Are you wanting to merge this in?

AxonC:
@CalumTowers yes, please!

PR #1208, opened 2018-07-22
CORE-119 - Tidy Up ACP

CalumTowers:
- Review sidebar, extract to individual files and ensure permissions are correct.
- Remove the hundreds of comments from AdminLTE that actually aren’t useful.
- Remove anything that is “bloaty” like the online indicator that serves little purpose.
- Improve consistency for yield vs content.
- Ensure scripts/stylesheets are properly included/extracted.
- Remove unused files / commented out blocks of code.

There is probably further work to do to bring the `visit-transfer` and `network-data` admin panels in with the rest of the code.


resources/views/adm/layout/scripts.blade.php
```diff
@@ -0,0 +1,66 @@
...
+        btnCancelClass: "btn btn-xs",
+        singleton: true,
+        onConfirm: function (event, element) {onConfirm(event, element)},
+    });
+
+    // override this function to implement
+    function onConfirm(event, element) {}
+
+</script>
+
```
atoff: Petty, but could combine these script blocks?

CalumTowers: 161200b

CalumTowers:
@atoff Sounds good. Would like to swap all of the `@if` to `@can` at some stage but I think that can wait.

PR #1212, opened 2018-07-24
CORE-148 Tidy Routes

CalumTowers:


CalumTowers:
@AxonC @atoff Can you both sanity check this please? Especially the fairly sizeable moving of VT adm name spacing and routing. Pretty sure I got everything...

PR #1215, opened 2018-07-27
CORE-120 - Non-Production Environment Notification

atoff:
image
image

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -179,5 +179,5 @@ function minutesToHours($minutes)
 
 function is_local_environment()
 {
-    return env('APP_IS_LOCAL') == 'true' || App::environment('staging');
+    return env('APP_IS_LOCAL') == 'true';
 }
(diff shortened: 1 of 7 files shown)


AxonC:
Could we have it above the navigation bar? I feel I'd notice that more.

CalumTowers:
Thinking about it, would we be better having this as a separate env variable?

CalumTowers:
You’re right. My thinking is it may become a hindrance for developing (seeing what it will actually look like)

My main aim for this is for alpha/beta testing, rather than local envs. 

I think we can kill two birds with one stone by setting this to only display if the env is staging?

atoff:
I think that sounds good to me. I may also implement a optional .env variable as well so that you can use it on dev as I know I would like it so I can see at a glance I am not on live! Will action tomorrow

app/helpers.php
```diff
@@ -176,3 +176,7 @@ function minutesToHours($minutes)
 
     return sprintf('%02.2d:%02.2d', floor($seconds / 3600), $seconds / 60 % 60);
 }
+
+function is_local_environment() {
+    return env("APP_IS_LOCAL") == "true" || App::environment('staging');
```
CalumTowers: A staging env isn't really local, we should probably restrict this to just return the env true/false

PR #1220, opened 2018-08-10
CORE-152 Use RefreshDatabase

CalumTowers:


CalumTowers:
@atoff @AxonC Losing the will with Travis erroring on things like Slack... Are we happy to merge?

PR #1224, opened 2018-08-11
CORE-147 - New Landing Page

CalumTowers:
- Moved mship management from "index" to "dashboard"
- Implemented a new `scss` for homepage
- Refactored airports work from @atoff into the Site namespace / site routing group

Homepage blade file is being kept away from GH for now [but it is in Google Drive](https://drive.google.com/file/d/1lElDn9N_UvtdxldB70hjvHIrqnrYMYPr/view?usp=sharing). There will be further amends to the homepage before going live (and perhaps style sheets) but want to merge this in so I have the basic routing structure etc.


app/Http/Controllers/Site/HomePageController.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace App\Http\Controllers\Site;
+
+class HomePageController extends \App\Http\Controllers\BaseController
+{
+    public function view()
```
AxonC: Could you use the __invoke() function on this controller, given it is a single action?

CalumTowers: Good spot!

resources/views/site/airport/index.blade.php
```diff
@@ -10,14 +10,14 @@
                         <div class="col-md-6 text-center">
                             <ul class="list-unstyled">
                                 @foreach($airports->first() as $airport)
-                                    <li>@if($airport->major)<strong>@endif<a href="{{route('airport.view', $airport->icao)}}">{{$airport->name}} ({{$airport->icao}} / {{$airport->iata}})</a>@if($airport->major)</strong>@endif</li>
+                                    <li>@if($airport->major)<strong>@endif<a href="{{route('site.airport.view', $airport->icao)}}">{{$airport->name}} ({{$airport->icao}} / {{$airport->iata}})</a>@if($airport->major)</strong>@endif</li>
```
AxonC: There isn't anything wrong with how we are doing this but I feel we could utilise a span and apply a css class (font-weight-bold) conditionally, rather than fanny around with the <strong> tag.

E.g. `<span class="@if ($airport->major) font-weight-bold @endif">`

CalumTowers: `font-weight-bold` is not available in bootstrap 3 and the accepted method is using `<strong>` (ref: https://getbootstrap.com/docs/3.3/css/)

AxonC: Serves me right for reading the v4 docs. Whoops!

PR #1225, opened 2018-08-13
CORE-108 - Students Position in Waiting List

AxonC:


app/Listeners/Training/WaitingListEventSubscriber.php
```diff
@@ -8,17 +8,22 @@ class WaitingListEventSubscriber
 {
     public function accountAdded($event)
     {
-        return Log::channel('stack')->info("Account {$event->account} ({$event->account->id}) was added to {$event->waitingList}");
+        return Log::channel('training')->info("Account {$event->account} ({$event->account->id}) was added to {$event->waitingList}");
     }
 
-    public function accountPromoted()
+    public function accountPromoted($event)
```
CalumTowers: All of the logs need to tell us who performed the action too

app/Events/Training/AccountRemovedFromWaitingList.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * Get the channels the event should broadcast on.
+     *
+     * @return \Illuminate\Broadcasting\Channel|array
+     */
+    public function broadcastOn()
+    {
+        return new PrivateChannel('channel-name');
```
CalumTowers: Now we're implementing logging properly, do we need to give this a channel name?

AxonC: This isn't for logging - broadcast is for if we were using Laravel Echo and real-time events

CalumTowers: Oops...

PR #1226, opened 2018-08-16
CORE-161 Improve Member CERT Validation

atoff:


CalumTowers:
Is it worth us doing proper validation on the data we get in for all fields to prevent this reoccurring with other bad data?

CalumTowers:
@atoff Just totally changed my mind! Think it makes more sense to have this into develop as an improvement and just snooze the bug reports. Its not causing and major issues so I think its fine to leave in prod for now!

Thanks 👍

PR #1228, opened 2018-08-18
CORE-166 Utilise ViewMake for all controllers

atoff:
Also, ensure that we get passed the account of semi-authed users


app/Http/Controllers/Mship/Feedback.php
```diff
@@ -70,7 +70,7 @@ public function getFeedback(Form $form)
             $question->form_html .= sprintf($question->type->code, $question->slug, old($question->slug));
         }
 
-        return view('mship.feedback.form', ['form' => $form, 'questions' => $questions]);
+        return $this->viewMake('mship.feedback.form', ['form' => $form, 'questions' => $questions]);
```
AxonC: Do you need to add ->with here?

CalumTowers:
@atoff May be worth doing a basic "visit as, assert successful" test for each of the pages we touch here? Verify that viewMake is not returning any errors

tests/Feature/SmartcarsTest.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Tests\Feature;
```
CalumTowers: Can we move this into its own namespace? Perhaps call the test something like `SmartcarsInterfaceTest` as, at some stage, I'd like to test the API

PR #1230, opened 2018-08-22
CORE-172 Install Nova

A-Lawrence:
Nova setup.

As Nova is a private set of files, depending on a license, the structure here allows for it to:

* Be part of the composer.json
* Avoid throwing errors about it missing for those without a license


app/Nova/User.php
```diff
@@ -0,0 +1,112 @@
...
+namespace App\Nova;
+
+use Laravel\Nova\Fields\ID;
+use Illuminate\Http\Request;
+use Laravel\Nova\Fields\Text;
+use Laravel\Nova\Fields\Gravatar;
+use Laravel\Nova\Fields\Password;
+
+/**
+ * @codeCoverageIgnore
```
CalumTowers: I think it would be more suitable to exclude the `/Nova/` folder recursively using the `codecov.yaml` file. Ref: https://docs.codecov.io/v4.3.6/docs/ignoring-paths

composer.json
```diff
@@ -1,108 +1,154 @@
...
+      "name": "Craig Milne"
     },
-    "config": {
-        "platform": {
-            "php": "7.2"
-        },
-        "preferred-install": "dist",
-        "sort-packages": true
+    {
+      "name": "Calum Axon"
```
CalumTowers: You're too used to working with me! Two L's for @AxonC

config/app.php
```diff
@@ -159,7 +159,7 @@
         Vatsim\Xml\XmlServiceProvider::class,
         Barryvdh\Debugbar\ServiceProvider::class,
         Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider::class,
-        Maknz\Slack\SlackServiceProvider::class,
+//        Maknz\Slack\SlackServiceProvider::class,
```
CalumTowers: Can you expand on the error that this was throwing that led to removing it?

app/Nova/User.php
A-Lawrence: I forgot you used this. I added these tags to exclude from code coverage in php unit. Will address tomorrow

composer.json
A-Lawrence: No. He shall only have one.

config/app.php
A-Lawrence: The slack service provider utilises application::share which was removed two (?) Laravel versions ago. Unsure if it's even still being used?

app/Nova/User.php
CalumTowers: Ah, makes sense

composer.json
CalumTowers: 🙄

config/app.php
CalumTowers: Doesn't appear to be used, can just remove the line

composer.json
AxonC: The distinction is very important! One of us can speak properly allegedly...

PR #1232, opened 2018-08-23
CORE-156 Teamspeak Registration Button

atoff:
Seems to work in combo with a temporary password


atoff:
@CalumTowers I have tested this in pieces, and it seems to work in conjunction with a temp password. However, I have been unable to get the TeamSpeak server running locally (it seems to never recognise the new registration)... Perhaps you should test and see if you are happy?

CalumTowers:
Thanks! I’ll take a look later. I’m planning to add TeamSpeak to the new Contributing Guide. Once you go through the setup faff in Vagrant, it’s relatively easy!

PR #1233, opened 2018-08-24
CORE-162 Stop trashed references appearing on admin

atoff:


CalumTowers:
@atoff Can you gander as to why Travis is failing?

PR #1239, opened 2018-08-25
CORE-167 FTE Navigation Checks

atoff:


resources/views/layout.blade.php
```diff
@@ -126,25 +126,23 @@
...
+                    <ul class="nav navbar-nav navcustom">
+                        <li class="dropdown dropdown-large">
+                            <a href="#" class="dropdown-toggle" data-toggle="dropdown">Pilots <b class="caret"></b></a>
+                            <ul class="dropdown-menu dropdown-menu-large row mainmenu_dropdown">
+                                <li class="col-sm-12">
+                                    <ul>
+                                        <li class="dropdown-header">Flight Training Exercises</li>
+                                        <li>{!! link_to_route('fte.dashboard', 'Dashboard') !!}</li>
+                                        <li>{!! link_to_route('fte.guide', 'Guide') !!}</li>
+                                        {{--<li>{!! link_to_route('fte.map', 'Map') !!}</li>--}}
```
AxonC: Do we still need this?

tests/Feature/SmartCars/SmartcarsWebInterfaceTest.php
```diff
@@ -33,10 +32,40 @@ public function itRedirectsFromDashboardAsGuest()
...
     {
         $this->actingAs($this->account, 'web')->get(route('fte.dashboard'))
-                ->assertSuccessful();
+                ->assertSuccessful()
+                ->assertSeeText('View All Exercises')
+                ->assertDontSee('No Flight Training Exercises are available at the moment.');
+    }
+
+    /** @test * */
+    public function itDoesNotShowExercisesButtonWhenNoneEnabled()
```
AxonC: I'd advise changing the conventions of these tests in inline with the other feature tests i.e. testCalumIsAngry

resources/views/layout.blade.php
atoff: @CalumTowers

resources/views/fte/dashboard.blade.php
```diff
@@ -73,7 +73,11 @@
                 <div class="panel" id="allExercises">
                     <div class="panel-body">
                         <div class="text-center">
-                            <a href="{{ route('fte.exercises') }}" class="btn btn-primary">View All Exercises &gt;&gt;</a>
+                            @if(\App\Models\Smartcars\Flight::enabled()->count() > 0)
```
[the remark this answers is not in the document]
atoff: Nope, because the `$exercises` variable only has featured exercises to be displayed on the dash. We could have enabled exercises but non featured...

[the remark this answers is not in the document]
CalumTowers: I think we would actually be better removing this then because we won't ever have a situation where we have no exercises available at all.

PR #1242, opened 2018-08-30
[WIP] CORE-159 - Waiting List Status

AxonC:


app/Models/Training/WaitingListAccount.php
```diff
@@ -65,4 +66,19 @@ public function incrementPosition($value = 1)
 
         return $this->position;
     }
+
+    public function atcHourCheck()
+    {
+        // for a user in a waiting list, they should have > 12 hours controlled within the UK.
+        $controllingSessions = Atc::whereAccountId($this->account_id)
+            ->whereBetween('disconnected_at', [\Carbon\Carbon::parse('3 months ago'), now()])->isUk();
```
atoff: I assume this might need a `->get()` or something here?

AxonC: It seems to work without this... could you pull and try this for me just to confirm?

atoff: Nevermind, looks like the `sum()` function also acts as a `get()`er

AxonC:
PR merged approved by @CalumTowers via phone call.

PR #1261, opened 2018-09-14
[WIP] CORE-189 - Waiting List View Students

AxonC:
Adds  ability to view students in a waiting list within Nova. Excludes position calculations as this is to be completed as part of CORE-190


app/Models/Training/WaitingList.php
```diff
@@ -18,10 +18,10 @@ class WaitingList extends Model
     const ATC_DEPARTMENT = 1;
     const PILOT_DEPARTMENT = 2;
 
-    public function getRouteKeyName()
-    {
-        return 'slug';
-    }
+//    public function getRouteKeyName()
```
CalumTowers: Should this be commented out?

app/Nova/WaitingList.php
```diff
@@ -51,6 +52,8 @@ public function fields(Request $request)
                 1 => 'ATC Training',
                 2 => 'Pilot Training',
             ])->displayUsingLabels()->rules(['required'])->sortable(),
+
```
CalumTowers: We need to account for the slug within this form somewhere, or give it a sensible default based on the name

package.json
```diff
@@ -1,30 +1,32 @@
...
+    "private": true,
+    "scripts": {
+        "dev": "npm run development",
+        "development": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "watch": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "watch-poll": "npm run watch -- --watch-poll",
+        "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "prod": "npm run production",
+        "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "build-waiting-lists-manager": "cd nova-components/WaitingListsManager && npm run dev",
```
CalumTowers: This appears to be failing when running `cannot find path specified events.js:137`

We also need to consider how we manage these moving forward. This could get _very_ messy if we just keep adding them here..

CalumTowers:
@AxonC What do we want to do about this at the moment?

PR #1273, opened 2018-09-17
Suggested use of collection methods...

A-Lawrence:
Suggested use of collection methods... (Please test this!)


app/Http/Controllers/Site/StaffPageController.php
```diff
@@ -33,9 +33,9 @@ private function getStaffPhotos()
         try {
             $ipboard = new Ipboard();
 
-            foreach ($teamPhotos->keys()->toArray() as $staff) {
-                $teamPhotos->put($staff, $ipboard->getMemberById($staff)->photoUrl);
-            }
+            return $teamPhotos->map(function($value, $key) use($ipboard) {
```
CalumTowers: If we return here, the return on line 42 can be removed

PR #1278, opened 2018-09-23
CORE-195 - Laravel 5.7 Upgrade

CalumTowers:


package.json
```diff
@@ -11,8 +11,8 @@
   },
   "devDependencies": {
     "admin-lte": "2.3.11",
-    "axios": "^0.16.2",
-    "bootstrap": "^3.3.7",
+    "axios": "^0.18",
+    "bootstrap": "^4.0.0",
```
AxonC: Ooh someone is brave... are we sure this is going to be as simple as a upgrade?

routes/web.php
```diff
@@ -7,7 +7,3 @@
 Route::group(['domain' => config('app.url'), 'middleware' => ['admin']], function () {
     require base_path('routes/web-admin.php');
 });
-
-Route::group(['domain' => 'vats.im'], function () {
```
AxonC: Why are we removing this?

package.json
CalumTowers: Tested and appears to be working fine (Laravel now ships with 4.0.0 out of the box). Feel free to have a test!

atoff: +1 with the bootstrap upgrade, that will break quite a lot. Off the top of my head, panels are now cards to start with

routes/web.php
CalumTowers: No longer hosted by VATSIM UK or Core, now its own service at https://vats.im/ and https://github.com/NFarrington/vatsim-url-shortener

PR #1284, opened 2018-10-02
CORE-206 - Partial Auth Error Fix

CalumTowers:


app/Http/Controllers/BaseController.php
```diff
@@ -30,7 +30,7 @@ class BaseController extends \Illuminate\Routing\Controller
     public function __construct()
     {
         $this->middleware(function ($request, $next) {
-            if (Auth::check() || Auth::guard('vatsim-sso')->check()) {
+            if (Auth::check() || Auth::guard('web')->check()) {
```
atoff: Same as below, when no guard is specified, it should default to web

app/helpers.php
```diff
@@ -184,5 +184,5 @@ function is_local_environment()
 
 function currentUserHasAuth()
 {
-    return Auth::check() || Auth::guard('vatsim-sso')->check();
+    return Auth::check() && Auth::guard('vatsim-sso')->check() && Auth::guard('web')->check();
```
atoff: `Auth::check()` should default  to using the default guard, which is `web` in our case

Also, users won't be authenticated in multiple guards, so the `||` operator is required. I.e we want to show the logout button to people who have been authed through CERT as well as those who have also passed secondary authentication.

PR #1289, opened 2018-10-13
CORE-207 - Existing Slack Registration Error Message

AxonC:
Improves error message displayed when an existing slack registration is found. Resolves CORE-207.

### DIFF
--- a/app/Http/Controllers/Slack/Registration.php
+++ b/app/Http/Controllers/Slack/Registration.php
@@ -20,9 +20,7 @@ class Registration extends \App\Http\Controllers\BaseController
      */
     public function getNew()
     {
-        try {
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Slack/Registration.php
```diff
@@ -19,7 +19,10 @@ class Registration extends \App\Http\Controllers\BaseController
      */
     public function getNew()
     {
-        $this->authorize('register-slack');
+        if (!$this->account->can('register-slack')) {
```
CalumTowers: The `register-slack` Gate has other reasons it could error (e.g. if you're an international member) so this error message could be mis-leading

PR #1290, opened 2018-10-13
CORE-212 - Session Expired Redirects

atoff:
Closes CORE-212


resources/views/components/nav.blade.php
```diff
@@ -67,7 +67,7 @@
                                 <li>{!! link_to_route("site.atc.bookings", "Bookings") !!}</li>
                             </ul>
                         </li>
-                        @if(currentUserHasAuth() && Auth::user()->qualificationAtc->isS1)
+                        @if(currentUserHasAuth() && Auth::user()->qualificationAtc && Auth::user()->qualificationAtc->isS1)
```
CalumTowers: This menu option shows for all to avoid confusion and a redirect is present in the controller to handle non-S1s trying to get in

app/Http/Controllers/Site/HomePageController.php
```diff
@@ -17,9 +17,13 @@ public function __invoke()
 
     private function nextEvent()
     {
-        $html = file_get_contents('https://cts.vatsim.uk/extras/next_event.php');
+        try {
```
CalumTowers: What sort of errors are you getting to result in the need for a try catch here?

resources/views/components/nav.blade.php
atoff: https://github.com/VATSIM-UK/core/blob/master/resources/views/components/nav.blade.php#L70 this is already present on prod, I simply added this to stop it error-ing some tests

app/Http/Controllers/Site/HomePageController.php
atoff: Mainly to do with SSL handshake timeout issues / dodgy connections. Have found that this caused phpunit to error on my tests a couple of times

CalumTowers: Would we be better returning something to avoid the homepage looking foobar if it does throw an exception?

resources/views/components/nav.blade.php
CalumTowers: There should never be an instance where we have someone with auth and without qualifications. Which tests were erroring?

app/Http/Controllers/Site/HomePageController.php
atoff: Returning null causes it to show no upcoming events which is fine

CalumTowers: So it does! I built it properly then ;)

resources/views/components/nav.blade.php
atoff: Uh my tests use a user without quals ;)

CalumTowers: The `Account::class` factory should spin up a state and qualification ideally

atoff: Do we want to remove that restriction on the nav aswell?

CalumTowers: Yeah we can do and just handle it in the controller

PR #1292, opened 2018-10-15
Fix some wayward local env errors.

A-Lawrence:
The default config settings for core prevent changing to utilising other services (i.e. redis) for sessions and cache storage.

I've modified this to allow overriding, but preserve defaults.


A-Lawrence:
@CalumTowers - we can add them to `.env.example`, however they're all defaulted and _only_ required if they should change.  Do you want the .env littered with variables that aren't required for default functionality?  I'll add to app.json though

CalumTowers:
> @CalumTowers - we can add them to `.env.example`, however they're all defaulted and _only_ required if they should change. Do you want the .env littered with variables that aren't required for default functionality? I'll add to app.json though

There's probably a wider job to be done to clean up `.env` and the `config/` files.
Go ahead and add to `app.json` and I'll merge in 👍

PR #1299, opened 2018-10-26
CORE-42 - Account Touched

atoff:
Closes CORE-42


CalumTowers:
@atoff Thank you! I think this PR highlights the need for us to get a beta version of some our services (I think CTS and Forum at the very least) setup and configured on a staging area so that we can push branches to there if we want to test with some "proper" data vs Heroku.

CalumTowers:
> @CalumTowers idk about codeclimate

Looks like we can make a simple change to stop the vast majority of its squawking. Could you rename the snake_case variables to be camelCase please? The rest of the issues it highlights are out of scope for this PR.

app/Jobs/Mship/SyncToCTS.php
```diff
@@ -0,0 +1,55 @@
...
+    private $account;
+
+    public function __construct(Account $account)
+    {
+        $this->account = $account;
+    }
+
+    public function handle()
+    {
+        $ctsDatabse = config('services.cts.database');
```
CalumTowers: There is a small typo on `$ctsDatabase` in this class.

app/Listeners/Mship/SyncSubscriber.php
```diff
@@ -0,0 +1,42 @@
...
+            // Prevent unnecessary executions
+            return;
+        }
+
+        \App\Jobs\Mship\SyncToCTS::dispatch($event->account);
+        \App\Jobs\Mship\SyncToHelpdesk::dispatch($event->account);
+        \App\Jobs\Mship\SyncToMoodle::dispatch($event->account);
+        \App\Jobs\Mship\SyncToForums::dispatch($event->account);
+
+        Log::debug($event->account->real_name.' was queued to sync to external services');
```
CalumTowers: It would be good to log the user's CID at this point too

app/Jobs/Mship/SyncToForums.php
```diff
@@ -0,0 +1,99 @@
...
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    private $account;
+
+    public function __construct(Account $account)
+    {
+        $this->account = $account;
+    }
+
+    public function handle()
```
CalumTowers: Should we consider putting a concern on the `Account` model that the account `HasForumAccount`?

app/Jobs/Mship/SyncToCTS.php
```diff
@@ -0,0 +1,55 @@
...
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    private $account;
+
+    public function __construct(Account $account)
+    {
+        $this->account = $account;
+    }
+
+    public function handle()
```
CalumTowers: Should we consider putting a concern on the `Account` model that the account `HasCtsAccount`?

app/Listeners/Mship/SyncSubscriber.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Register the listeners for the subscriber.
+     *
+     * @param  \Illuminate\Events\Dispatcher $events
+     */
+    public function subscribe($events)
+    {
+        $events->listen(
+                \App\Events\Mship\AccountAltered::class,
+                '\App\Listeners\Mship\SyncSubscriber@syncToAllServices'
```
CalumTowers: Do we need to define this or can we just register the listener within the service provider to fire on the `AccountAltered` event?

atoff: This method is better for the future if we want to only sync certain services for certain events, which I could imagine happening at some point?

atoff:
> As a side note, should we consider deprecating the sync commands?

I wouldn't want to remove these entirely; the main reason I kept them was in case when this goes live and doesn't work we have a fall back. If your testing with the systems indicates its fine, I will update the command to fire a job instead?

PR #1325, opened 2018-11-03
CORE-229 - Intended Redirect Fix

CalumTowers:
Users that do not have a secondary password (and therefore only ever need to pass through the vatsim-sso guard) are being redirected to the homepage with every login, rather than to their intended route.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -111,6 +111,7 @@ class LoginController extends BaseController
             Auth::login(Auth::guard('vatsim-sso')->user(), true);
 
             $intended = Session::get('url.intended') ?? route('site.home');
+
             return redirect($intended);
         }
     }
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -110,7 +110,8 @@ class LoginController extends BaseController
         } else {
             Auth::login(Auth::guard('vatsim-sso')->user(), true);
 
-            return redirect('/');
+            $intended = Session::get('url.intended') ?? route('site.home');
+            return redirect($intended);
         }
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Auth/LoginController.php
```diff
@@ -110,7 +110,9 @@ protected function attemptSecondaryAuth()
         } else {
             Auth::login(Auth::guard('vatsim-sso')->user(), true);
 
-            return redirect('/');
+            $intended = Session::get('url.intended') ?? route('site.home');
```
atoff:
```suggestion
            $intended = Session::pull('url.intended', route('site.home'));
```

Also suggest moving this up to the top of the method and removing both existing `$intended` variables because its always used. The above also eliminates the need for the `Session::remove`

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -95,23 +95,16 @@ protected function setVatsimAuth($userId)
...
-        if (Session::has('url.intended')) {
-            $intended = Session::get('url.intended');
-            if (starts_with($intended, route('password.request'))) {
-                Session::remove('url.intended');
-
-                return redirect($intended);
-            }
-        }
-
+        $intended = Session::pull('url.intended', route('site.home'));
```
atoff: This needs to be after line 102, otherwise the next time they come around after secondary pass, the intended destination will already have been removed from the session.

atoff: I'm not quite sure you understood what I meant here - by having the `$intended` where it is in the code, a user who has to attempt secondary authentication (i.e gets taken to put in their secondary password), will not be redirected to the intended page. I was simply saying to move the pull of the intended url to below the `hasPassword` check so that this doesn't happen

PR #1338, opened 2018-11-12
CORE-226 - Application Counting

atoff:
Closes CORE-226

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,6 @@
 APP_ENV=development
 APP_DEBUG=true
+## Enable the Debug Bar, visible at the bottom of the page
 APP_DEBUGBAR=true
 APP_KEY=
 APP_URL=vukcore.test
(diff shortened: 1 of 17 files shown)


tests/Unit/VisitTransfer/ReferenceTest.php
```diff
@@ -64,4 +66,37 @@ public function itCanRejectItself()
...
+
+        // Create some references
+
+        factory(Reference::class, 20)->create([
+            'status' => function () use ($referenceTypes) {
+                return Base::randomElement(Base::randomElement($referenceTypes));
+            },
+        ]);
+
+        // Test
```
AxonC: Do we perhaps need to assert that the Cache::remember method is actually called when we try and get it from the model?

atoff: I mean its not essential for it to cache it - the fact that it is returning the data correctly shows that it works so imo its fine. If you are super keen on it I can put it in though

PR #1360, opened 2018-12-13
CORE-99 - Manage Waiting Lists

AxonC:


tests/Feature/Training/WaitingListFeatureTest.php
```diff
@@ -74,29 +76,37 @@ public function testAStudentCanOnlyBeInAListOnce()
...
+        $this->waitingList->addToWaitingList($account, $this->privacc);
+        $this->waitingList->addToWaitingList($account2, $this->privacc);
+        $this->waitingList->addToWaitingList($account3, $this->privacc);
+
+        $this->actingAs($this->privacc)->post("nova-vendor/waiting-lists-manager/accounts/{$this->waitingList->id}/demote", [
             'account_id' => $account->id,
-            'position' => 1,
-        ])->assertRedirect(route('training.waitingList.show', $this->waitingList))
-            ->assertSessionHas('success', 'Waiting list positions changed.');
+        ])->assertSuccessful();
```
CalumTowers: Should we assert that an event was dispatched here?

AxonC: Is that should we in the sense that you don't think its required, or should we in that you think we haven't already?

CalumTowers: @AxonC We haven't already?

tests/Feature/Training/WaitingListFeatureTest.php
```diff
@@ -74,29 +76,37 @@ public function testAStudentCanOnlyBeInAListOnce()
...
 
         Event::assertDispatched(AccountDemotedInWaitingList::class, function ($event) use ($account) {
```
AxonC: @CalumTowers *ahem*

PR #1361, opened 2018-12-13
CORE-299 - TeamSpeak Server Group Exception

atoff:
Closes CORE-299


atoff:
@CalumTowers not quite sure how best we can fix the root cause. Hopefully we can identify the user this effects through the logs unless you have any ideas? The logic seems fine for me

PR #1363, opened 2018-12-16
CORE-307 - Include MAN Into UK Network Data

CalumTowers:


app/Models/NetworkData/Atc.php
```diff
@@ -187,6 +187,7 @@ public static function scopeIsUK($query)
                 ->orWhere('callsign', 'LIKE', "STC\_%")
                 ->orWhere('callsign', 'LIKE', "LON\_%")
                 ->orWhere('callsign', 'LIKE', "LTC\_%")
+                ->orWhere('callsign', 'LIKE', 'MAN\_%')
                 ->orWhere('callsign', 'LIKE', 'EGGX%')
                 ->orWhere('callsign', 'LIKE', 'EGTT%')
                 ->orWhere('callsign', 'LIKE', 'EGPX%');
```
AxonC:
```suggestion
                ->orWhere('callsign', 'LIKE', 'EGPX%');
                ->orWhere('callsign', 'LIKE', 'LXGB_%');
                ->orWhere('callsign', 'LIKE', 'LCRA_%');
                ->orWhere('callsign', 'LIKE', 'FHAW_%');
```

AxonC: I noticed when I did the manual queries that we were missing some of the minor positions. Might as well add them whilst we are at it!

app/Models/NetworkData/Atc.php
```diff
@@ -187,9 +187,16 @@ public static function scopeIsUK($query)
                 ->orWhere('callsign', 'LIKE', "STC\_%")
                 ->orWhere('callsign', 'LIKE', "LON\_%")
                 ->orWhere('callsign', 'LIKE', "LTC\_%")
+                ->orWhere('callsign', 'LIKE', 'MAN\_%')
                 ->orWhere('callsign', 'LIKE', 'EGGX%')
                 ->orWhere('callsign', 'LIKE', 'EGTT%')
-                ->orWhere('callsign', 'LIKE', 'EGPX%');
+                ->orWhere('callsign', 'LIKE', 'EGPX%')
```
atoff: This shouldn't be necessary due to the `EG%` rule above

PR #1367, opened 2018-12-22
Add Pretty PHPUnit

CalumTowers:
Add `codedungeon/phpunit-result-printer` to improve the look of PHPUnit.

Before:
before

After:
after


CalumTowers:
@AxonC One of the reasons I didn't submit this PR whilst on Windows! 😉

PR #1369, opened 2018-12-25
CORE-297 - Notification Handling & Authentication

atoff:
Closes CORE-297


app/Models/Mship/Concerns/HasNotifications.php
```diff
@@ -22,6 +22,15 @@ public function readSystemNotifications()
             ->withPivot(['created_at']);
     }
 
+    /**
+     * @param Notification $notification
+     * @return bool
+     */
+    public function hasReadNotification(SysNotification $notification)
```
CalumTowers: It would be good to add some assertions in the tests about this relationship.

e.g.

`$this->assertFalse($this->account->hasReadNotification($notification);`

`$this->post......`

`$this->assertTrue($this->account->hasReadNotification($notification);`

It may feel more natural for this to be done at unit level.

tests/Feature/Mship/NotificationTest.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+    namespace Tests\Feature\Mship;
```
CalumTowers: Not sure why StyleCI isn't quite picking it up but there is a fair amount of indentation on this file and the Factory?

PR #1380, opened 2018-12-28
CORE-332 - Revised Navigation

CalumTowers:


CalumTowers:
@atoff @AxonC Shall we deploy this change to the homepage and revisit the underlying nav later?

resources/views/site/home.blade.php
```diff
@@ -105,71 +106,109 @@ function toggleActive () {
...
+                        <li class="nav-item">
+                            <a class="nav-link" href="{{ route('site.marketing.live-streaming') }}">Live Streams</a>
+                        </li>
+                    </ul>
+                </li>
+                <li class="nav-item">
+                    <a class="nav-link" href="{{ route('mship.feedback.new') }}">Feedback</a>
+                </li>
+                <li class="nav-item">
+                    <a class="nav-link" href="#">Contact Us <span class="arrow"></span></a>
```
AxonC: Does this need to be a dropdown? Surely we can just direct them to the Helpdesk straight from the home page.

resources/views/site/home.blade.php
```diff
@@ -401,9 +453,6 @@ function toggleActive () {
 
 </script>
 <script src="{{ mix('js/home.js') }}"></script>
-@if(Carbon\Carbon::now()->month == 12 || Carbon\Carbon::now()->dayOfYear < 10)
```
AxonC: Are we getting rid of this?

resources/views/site/home.blade.php
```diff
@@ -105,71 +106,109 @@ function toggleActive () {
...
-                    <li class="nav-item">
-                        <a class="nav-link text-white" href="https://twitter.com/vatsimuk" target="_blank"><i class="fab fa-twitter"></i></a>
-                    </li>
-                    @if(currentUserHasAuth())
+                <li class="nav-item">
+                    <a class="nav-link active" href="#">Home <span class="arrow"></span></a>
+                    <ul class="nav">
                         <li class="nav-item">
-                            <a href="{{ route('login') }}" class="nav-link text-white">{{ $_account->full_name }} <i class="fas fa-user"></i></a>
+                            <a class="nav-link" href="#">About VATSIM UK</a>
```
atoff: `#`

resources/views/site/home.blade.php
atoff: I think we should keep it seeing as its defiantly going to be used again next year...

CalumTowers: The implementation of the original snow (by me) was fairly quick and dirty. Pretty confident we could do something better.
Also fairly certain we will have a refreshed design by then... I hope!

CalumTowers:
> is it possible to put a link to our CoC/T&Cs? There is no prominent link to this anywhere that I can see

We have a link to Policy and Regulations (that links to all policies on the Downloads). I envisage us having some kind of policy system at some point that forces users to agree to things so I'm content with leaving it as is if you are?

atoff:
> > is it possible to put a link to our CoC/T&Cs? There is no prominent link to this anywhere that I can see
> 
> We have a link to Policy and Regulations (that links to all policies on the Downloads). I envisage us having some kind of policy system at some point that forces users to agree to things so I'm content with leaving it as is if you are?

Not sure if I am being blind, but there is no link to that page under the  policies & regs stuff (probably because it isn't a download as it is in core). I agree we will have a policy 'agreeing' thing in the future but even then we will need it linked somewhere?

CalumTowers:
> > > is it possible to put a link to our CoC/T&Cs? There is no prominent link to this anywhere that I can see
> > 
> > 
> > We have a link to Policy and Regulations (that links to all policies on the Downloads). I envisage us having some kind of policy system at some point that forces users to agree to things so I'm content with leaving it as is if you are?
> 
> Not sure if I am being blind, but there is no link to that page under the policies & regs stuff (probably because it isn't a download as it is in core). I agree we will have a policy 'agreeing' thing in the future but even then we will need it linked somewhere?

Its under "Controllers" at the moment on this branch, where would you propose we place it?

atoff:
> > > > is it possible to put a link to our CoC/T&Cs? There is no prominent link to this anywhere that I can see
> > > 
> > > 
> > > We have a link to Policy and Regulations (that links to all policies on the Downloads). I envisage us having some kind of policy system at some point that forces users to agree to things so I'm content with leaving it as is if you are?
> > 
> > 
> > Not sure if I am being blind, but there is no link to that page under the policies & regs stuff (probably because it isn't a download as it is in core). I agree we will have a policy 'agreeing' thing in the future but even then we will need it linked somewhere?
> 
> Its under "Controllers" at the moment on this branch, where would you propose we place it?

I still can't find it? Just to be clear I am talking about https://www.vatsim.uk/community/terms-and-conditions, and it should probably go under the main home sub menu?

CalumTowers:
> > > > > is it possible to put a link to our CoC/T&Cs? There is no prominent link to this anywhere that I can see
> > > > 
> > > > 
> > > > We have a link to Policy and Regulations (that links to all policies on the Downloads). I envisage us having some kind of policy system at some point that forces users to agree to things so I'm content with leaving it as is if you are?
> > > 
> > > 
> > > Not sure if I am being blind, but there is no link to that page under the policies & regs stuff (probably because it isn't a download as it is in core). I agree we will have a policy 'agreeing' thing in the future but even then we will need it linked somewhere?
> > 
> > 
> > Its under "Controllers" at the moment on this branch, where would you propose we place it?
> 
> I still can't find it? Just to be clear I am talking about https://www.vatsim.uk/community/terms-and-conditions, and it should probably go under the main home sub menu (near policies) or community?

Added to the policies link! 👍

PR #1381, opened 2018-12-28
CORE-321 - Slack Regression Issues

CalumTowers:
Gate removal has been causing a few issues...

- Conditional on membership dashboard always showing as not eligible.
- Email invite never being sent.

If someone doesn't have a `slack_id`, they should be eligible. Simple!

### DIFF
--- a/app/Notifications/Mship/SlackInvitation.php
+++ b/app/Notifications/Mship/SlackInvitation.php
@@ -3,7 +3,6 @@
 namespace App\Notifications\Mship;
 
 use App\Notifications\Notification;
-use Gate;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Notifications\Messages\MailMessage;
@@ -30,7 +29,7 @@ class SlackInvitation extends Notification implements ShouldQueue
      */
     public function via($notifiable)
     {
-        if (Gate::forUser($notifiable)->allows('register-slack')) {
+        if (!$notifiable->slack_id) {
             return ['mail', 'database'];
         } else {
             return [];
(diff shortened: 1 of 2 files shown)


atoff:
I am asuming this means that slack is now meant to be open to all VATSIM members and not just UK/V/T members?

CalumTowers:
> I am asuming this means that slack is now meant to be open to all VATSIM members and not just UK/V/T members?

Yep, we opened it up in 432036303de93bf2ae0740d18f647b7c47592c78

app/Notifications/Mship/SlackInvitation.php
```diff
@@ -30,7 +29,7 @@ public function __construct()
      */
     public function via($notifiable)
     {
-        if (Gate::forUser($notifiable)->allows('register-slack')) {
+        if (!$notifiable->slack_id) {
```
AxonC:
```suggestion
         return !$notifiable->slack_id ? ['mail', 'database'] : [];
```
Not a big one but might make it a bit simpler? 6 of one, half a dozen of the other on this one.

CalumTowers: If I look at your code, I recognise it as "proper".
But when I glance at both, I instantly know what mine is doing. There is a degree of "mental debt" being added by decoding the `!` then the `?` etc with your suggestion.

Maybe I'm just not very good! 😉

AxonC: I hear ya! To me, its good to look at because I know left is true and right is false. I'm not going to block it based on this though :laughing:

CalumTowers: @AxonC Good, get it approved 💯

PR #1382, opened 2018-12-28
CORE-309 - Initial UKCP API setup

CalumTowers:


app/Libraries/UKCP.php
```diff
@@ -0,0 +1,18 @@
...
+
+    /** @var string */
+    private $apiKey;
+
+    /**
+     * UKCP constructor.
+     */
+    public function __construct()
+    {
+        $this->apiKey = env('UKCP_KEY');
```
AxonC: Since we cache the env variable in prod (I assume), should we extract this to a config file: REF: https://laravel.com/docs/5.7/configuration#configuration-caching

CalumTowers: 1f67400

PR #1384, opened 2018-12-28
CORE-312 - Delete UKCP Tokens

CalumTowers:


app/Http/Controllers/Mship/UKCP.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Support\Facades\Redirect;
+
+class UKCP extends BaseController
+{
+    /** @var UKCP */
+    private $ukcp;
+
+    public function __construct()
+    {
+        $this->ukcp = new UKCPLibrary;
```
AxonC: Should we be resolving this from the service container via dependency injection (via the constructor)?  I'm always a bit cautious of using composition in this way in controllers. Again, just a personal preference and it if it works, don't fix it ;)

CalumTowers: 1f75166

PR #1389, opened 2018-12-31
CORE-323 - Network Data Indices

CalumTowers:
Resolves CORE-323


database/migrations/2018_12_31_122052_add_network_data_indices.php
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('networkdata_atc', function (Blueprint $table) {
+            $table->dropIndex('callsign_index');
```
atoff: This `dropIndex` needs a table prefixed as well?

CalumTowers: Oops! c7a74dc

PR #1395, opened 2019-01-02
CORE-280 - Viewing FTEs

AxonC:


app/Nova/Pirep.php
```diff
@@ -0,0 +1,133 @@
...
+     * @param  \Illuminate\Http\Request  $request
+     * @return array
+     */
+    public function fields(Request $request)
+    {
+        return [
+            ID::make()->sortable(),
+
+            BelongsTo::make('Bid', 'bid', 'App\Nova\Bid'),
+
```
atoff: Should have the "exercise" (flight) displayed somewhere here (and displayed on index)

app/Models/Smartcars/Pirep.php
```diff
@@ -59,6 +59,12 @@ public function bid()
         return $this->belongsTo(\App\Models\Smartcars\Bid::class, 'bid_id', 'id');
     }
 
+    public function account()
+    {
+        return $this->hasManyThrough(\App\Models\Mship\Account::class, \App\Models\Smartcars\Bid::class,
```
CalumTowers: I know you're following the trend of the model but can we import all of the classes being referenced as use statements please?

app/Nova/Pirep.php
```diff
@@ -0,0 +1,133 @@
...
+            Text::make('Pilot', function () {
+                return $this->account->first()->name;
```
atoff: This causes a property of non-object exception on updating

app/Nova/Pirep.php
```diff
@@ -0,0 +1,133 @@
...
+            Text::make('Flight Time'),
+
+            new Panel('Log & Comments', [
+                Textarea::make('Comments'),
+                Textarea::make('Log'),
+            ]),
+
+            new Panel('Approval Information', [
+                Boolean::make('Passed'),
+                Text::make('Pass Reason')->onlyOnDetail()->canSee(function () {
```
CalumTowers: These db fields are poorly named. Pass reason actually returns "Passed" or it'll return the failed reason (e.g. "you failed because you went outside of the altitude restriction at posrep X").

So this only needs to show with a failed_at

AxonC:
> I feel like the bid shouldn't even be a thing in nova - it isn't in core adm at the mo, and its seems to me like it a bit of a waste of space? Would suggest hiding from sidebar. If we are keeping the bid, it should at least have the pilot & prirep associated with it in the detail view, and it should move under smartcars in the sidebar?
> 
> Also, is toggling/editing meant to be working in this pirep? I get an error when trying to update and setting passed to false - see specific line below.
> 
> Also - fix the tests :))

This PR is only concerned with _viewing_ for now - other PRs will address editing

PR #1399, opened 2019-01-09
CORE-295 - Fix Teamspeak Exception

atoff:
Closes CORE-295


app/Console/Commands/TeamSpeak/TeamSpeakDaemon.php
```diff
@@ -47,7 +47,13 @@ public function handle()
             try {
                 self::$connection->getAdapter()->wait();
             } catch (TeamSpeak3_Transport_Exception $e) {
-                self::$connection = $this->establishConnection();
+                try {
+                    self::$connection = $this->establishConnection();
+                } catch (TeamSpeak3_Transport_Exception $e) {
+                    // Connection failed, let the loop restart and try again
+                    $this->log('TeamSpeak connection failed: '.$e->getMessage(). '. Trying again in 15 seconds...');
```
CalumTowers: Are we potentially going to miss this if it isn't working for a few days? Do we need to try three times and then throw an exception or something?

atoff: Do you mean something like I just pushed?

CalumTowers:
@atoff Can we amend to camelCase before we merge please?

app/Console/Commands/TeamSpeak/TeamSpeakDaemon.php
```diff
@@ -41,13 +41,26 @@ class TeamSpeakDaemon extends TeamSpeakCommand
     public function handle()
     {
         self::$connection = $this->establishConnection();
+        $connection_failures = 0;
```
CalumTowers: Needs amending!

atoff: That's what I get for using the Git IDE :/

PR #1400, opened 2019-01-10
CORE-304 - Fix Slack-Core Interaction

atoff:
- Removes old references
- Implements a more robust internal Slack API for messaging
- Makes existing implementations use it
image
image
image

When this is deployed: Have to make sure the bot is invited/in #web_alerts, and that the token is set in the .env (should already be)

Closes CORE-304


atoff:
Yeah the main thing with those packages you suggested is that they use a webhook as opposed to a App integration. Looking to the future, we could do a lot more using a bot, and one of the reasons I think we currently use this package is because webhooks can't get details about users, just send messages.

Looking at the Laravel Notifications, it seems that much of what I have implemented is available there (and they also use the bot). Pretty much everything we do with slack at the moment could be done in the notifications, bar the slack manager itself which will still need the current package. 

Personally, I think we should keep the current package, and refactor some of the main messages into notifications

CalumTowers:
> Personally, I think we should keep the current package, and refactor some of the main messages into notifications

Yeah I agree. It would be good to play around with some of the messages and turn them into notifications just to get a feel for what it could look like?

atoff:
> * Get all users
> * Get a user

As far as I know, these aren't supported by webhooks

CalumTowers:
> > * Get all users
> > * Get a user
> 
> As far as I know, these aren't supported by webhooks, and you therefore need a bot-oriented package

I'm talking about the idea of making a wrapper around the API without a package; e.g.

https://api.slack.com/methods/users.list
https://api.slack.com/methods/users.info

atoff:
> I'm talking about the idea of making a wrapper around the API without a package; e.g.
Oh - We could definitely do that, but I'm not sure if it's worth spending time on it when the package does this and allows for expansion to other methods? I mean we could still wrap it, but I personally don't see much benefit in writing all of the logic that would be required to lose the package.

PR #1404, opened 2019-01-13
Publish UKCP

CalumTowers:


app/Providers/AuthServiceProvider.php
```diff
@@ -64,28 +64,7 @@ public function boot()
         });
 
         Gate::define('ukcp-beta', function ($user) {
```
AxonC: Do you want to just rename this to uckp now?

PR #1419, opened 2019-02-02
CORE-338 - Refactor FTE Image Processing

AxonC:
Refactoring the FTE image upload and retrieve process.


AxonC:
Obviously this brings about a breaking change in how images are stored and retrieved. It might be that we have to manually enter the file names of the images in production.

PR #1421, opened 2019-02-09
Change of Marketing Director

JoshHowker:
As per - https://community.vatsim.uk/news/united-kingdom/new-marketing-director-r246/  

Spoken to Loui and this is how he wants it formatting with the removal of Social Media Manager, until Marketing update has been posted.

### DIFF
--- a/resources/views/site/staff.blade.php
+++ b/resources/views/site/staff.blade.php
@@ -96,18 +96,10 @@
                     </div>
                     <div class="panel-body">
                         <h4>Marketing Director (VATUK4)</h4><br/>
-                        <img src="{{ $teamPhotos[5161] }}"
+                        <img src="{{ $teamPhotos[6738] }}"
                              width=50px
                              class="img-responsive center-block profile-picture"/>
-                        <p class="text-center">Trevor Hannant</p>
-
-                        <h4>Marketing Team</h4>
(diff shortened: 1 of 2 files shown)


resources/views/site/staff.blade.php
```diff
@@ -96,18 +96,10 @@ class="img-responsive center-block profile-picture"/>
                     </div>
                     <div class="panel-body">
                         <h4>Marketing Director (VATUK4)</h4><br/>
-                        <img src="{{ $teamPhotos[5161] }}"
+                        <img src="{{ $teamPhotos[6738] }}"
```
AxonC: Unfortunately, it isn't quite this simple! The only other change you need is on line 29 of StaffPageController. Remove 5161 and replace with 6738 and it will prepare the image correctly, and allow our test suite to pass. Let me know if you need a hand.

AxonC: REF: https://github.com/VATSIM-UK/core/blob/82ccc60431a166ad75a9f0df9c6b816cda8dd84a/app/Http/Controllers/Site/StaffPageController.php#L29

(I would make the changes myself, but it is from a forked branch)

AxonC:
Thanks @JoshHowker!

PR #1427, opened 2019-02-10
CORE-259 - Implement Facility Open

atoff:
Note: on prod, pretty sure most if not all facilities are currently not open, will need to be rectified before deploy

Closes CORE-259


app/Http/Requests/VisitTransfer/FacilityCreateUpdateRequest.php
```diff
@@ -41,6 +42,8 @@ public function messages()
         return [
             'name.required' => 'You must provide a name of at least 5 characters.',
             'name.min' => 'You must provide a name of at least 5 characters.',
+            'open.required' => 'You must provide whether or not the facility is open for applications.',
```
CalumTowers: Not sure this makes sense (grammatically)

PR #1443, opened 2019-03-13
Add mship dashboard link to adm dropdown

inventor02:


CalumTowers:
Thanks @inventor02 but I don't want to share too much nav between the admin panel and frontend. We'll shortly be decommissioning this admin panel too.

PR #1449, opened 2019-03-27
CORE-336 - Events on Booking Sidebar

atoff:
image

Without event:

image

Closes CORE-336


atoff:
@daviesnathan are you happy if I just change the colour of this to the same as the button, and leave it like this (we can always change later!)

image

CalumTowers:
@daviesnathan @atoff Who is working on this at the moment?

atoff:
@CalumTowers I think this is probably good to go. It has been sitting here for too long, it was originally only a question of the colour for the event but I think I have found an acceptable compromise above.

CalumTowers:
> @CalumTowers I think this is probably good to go. It has been sitting here for too long, it was originally only a question of the colour for the event but I think I have found an acceptable compromise above.

Agreed! Can you resolve conflicts and then merge? :)

PR #1456, opened 2019-06-03
CORE-368 - Null Referee Bug

atoff:
Closes CORE-368


app/Models/Mship/Account.php
```diff
@@ -257,6 +257,11 @@ public static function findOrRetrieve($accountId)
 
             $account = self::find($accountId);
 
+            if (!$account) {
+                // User doesn't exist at VATSIM.NET
+                return false;
```
AxonC: In accordance with the rest of the flows within this function, would it not be more germane to throw an exception here instead?

atoff: Yeah fair point, I had been thinking about doing that

PR #1457, opened 2019-06-03
CORE-370 - Admin Member Sync

atoff:
Closes CORE-370


app/Http/Controllers/Adm/Mship/Account.php
```diff
@@ -516,4 +518,17 @@ public function postImpersonate(Request $request, AccountData $mshipAccount)
         return Redirect::to(URL::route('mship.manage.dashboard'))
             ->withSuccess('You are now impersonating this user - your reason has been logged. Be good!');
     }
+
+    public function sync(AccountData $mshipAccount)
```
AxonC: Can we maybe switch this out to a more useful controller? Just so we don't bloat out this already bloated controller more.

PR #1462, opened 2019-06-12
CORE-380 - Obsolete Code Removal

atoff:
Closes CORE-380


atoff:
@CalumTowers @AxonC anything else you can think of not covered by these changes?

PR #1467, opened 2019-06-26
CORE-395 - Waiting List Notes - Endpoint for editing notes

AxonC:
Methods may look similar but the future will involve firing events and logging any changes in notes and their contents.


app/Models/Training/WaitingListAccount.php
```diff
@@ -165,4 +165,13 @@ public function addNote($contents)
 
         return $this->notes;
     }
+
+    public function editNote($contents)
```
atoff: Should this perhaps be renamed `editNotes` in line with the field name? Also, is there a specific reason why we would want this as a function rather than just a set mutator? At the moment, might just be for me, but when I see `addNote` and `editNote` it makes me think that it is a cumulative add (when it might not be?)

tests/Unit/Training/WaitingListAccountTest.php
```diff
@@ -213,4 +213,19 @@ public function itCanHaveNotesAdded()
 
         $this->assertEquals('This is a note', $waitingListAccount->fresh()->notes);
     }
+
+    /** @test */
+    public function itCanHaveNotesModified()
+    {
+        $account = factory(Account::class)->create();
```
atoff: I'm not sure how up to date the epic is with master, but in line with the testing housekeeping I did, `account` should become `user`, and if you are just creating a bog standard user you should use `$this->user` (https://github.com/VATSIM-UK/core/pull/1461/files#diff-d4c8c6dc8769324fc27cfdac19f05cafR50). It might just be a case of doing the petty renaming of stuff when we come to merge nova into master, but up to you!

AxonC: Yeah, I'd rather do that when we get the feature/epic done if that's alright?

app/Models/Training/WaitingListAccount.php
AxonC: New requirements dictate that we aren't going to distinguish between editing and creating a note, so expect a change shortly.

PR #1469, opened 2019-07-02
CORE-400 - Waiting List Notes - Modification Logging

AxonC:


app/Listeners/Training/WaitingList/LogNoteChanged.php
```diff
@@ -0,0 +1,34 @@
...
+use Illuminate\Support\Facades\Log;
+
+class LogNoteChanged
+{
+    /**
+     * Create the event listener.
+     *
+     * @return void
+     */
+    public function __construct()
```
atoff: Can we get rid of this empty construct?

PR #1471, opened 2019-07-03
CORE-358 - Endorsement Checker Updates

atoff:
Closes CORE-358


app/Http/Controllers/Adm/Atc/Endorsement.php
```diff
@@ -7,61 +7,42 @@
...
+        $endorsements = EndorsementModel::get(['id', 'name']);
 
         return $this->viewMake('adm.atc.endorsement.index')
                   ->with('endorsements', $endorsements);
     }
 
     public function getUserEndorsement(Request $request)
     {
-        $requirements = EndorsementModel::where('endorsement', $request->input('endorsement'))->get();
+        $endorsement = EndorsementModel::with('conditions')->find($request->input('id'))->first();
```
AxonC: Could you not use findOrFail here and then catch the exception?

app/Http/Controllers/Adm/Atc/Endorsement.php
```diff
@@ -7,61 +7,42 @@
...
-use Validator;
+use Illuminate\Support\Facades\Validator;
 
 class Endorsement extends AdmController
 {
     public function getIndex(Request $request)
     {
         $validator = Validator::make($request->all(), [
-          'endorsement' => 'required',
+          'id' => 'required',
```
AxonC: Is it worth validating the type of this?

app/Models/Atc/Endorsement.php
```diff
@@ -2,16 +2,32 @@
...
+
+    public function conditionsMetForUser(Account $user)
+    {
+        if ($this->allMet) {
+            return $this->allMet;
+        }
+
+        return $this->allMet = $this->conditions->filter(function ($condition) use ($user) {
+            return !$condition->isMetForUser($user);
+        })->count() == 0;
```
AxonC:
```suggestion
        })->isEmpty();
```
REF: https://laravel.com/docs/5.7/collections#method-isempty

PR #1479, opened 2019-07-21
CORE-411 - Teamspeak Controller Idle Exemption

atoff:
Closes CORE-411

### DIFF
--- a/app/Libraries/TeamSpeak.php
+++ b/app/Libraries/TeamSpeak.php
@@ -376,7 +376,7 @@ class TeamSpeak
     {
         $idleTime = floor($client['client_idle_time'] / 1000 / 60); // minutes
 
-        if ($member->hasPermissionTo('teamspeak/idle/permanent')) {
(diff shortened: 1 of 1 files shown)


app/Libraries/TeamSpeak.php
```diff
@@ -376,7 +376,7 @@ public static function checkClientIdleTime(TeamSpeak3_Node_Client $client, Accou
     {
         $idleTime = floor($client['client_idle_time'] / 1000 / 60); // minutes
 
-        if ($member->hasPermissionTo('teamspeak/idle/permanent')) {
```
AxonC: Can we go further with this check? At present, they could not be on a UK position.

atoff: I can't find anything to suggest this? It should just check if they have an active ATC session regardless of where it is?

AxonC: I was just taking that from the JIRA story

atoff: Ah wording subtleties... I will amend to user ;)

PR #1484, opened 2019-07-23
CORE-409 - Duplicate Names on TS

atoff:
Closes CORE-409

### DIFF
--- a/app/Libraries/TeamSpeak.php
+++ b/app/Libraries/TeamSpeak.php
@@ -280,10 +280,10 @@ class TeamSpeak
             $hasGracePeriod = Cache::has(self::CACHE_NICKNAME_PARTIALLY_CORRECT_GRACE.$client['client_database_id']);
 
             // Check to see if their name is at least partially right
-            if($member->isPartiallyValidDisplayName($client['client_nickname'])){
+            if ($member->isPartiallyValidDisplayName($client['client_nickname'])) {
 
                 // Check if they have been recently told but the grace has elapsed
-                if($recentlyTold && !$hasGracePeriod){
+                if ($recentlyTold && !$hasGracePeriod) {
                     self::pokeClient($client, trans('teamspeak.nickname.invalid.poke1'));
                     self::pokeClient($client, trans('teamspeak.nickname.invalid.poke2'));
                     self::kickClient($client, trans('teamspeak.nickname.invalid.kick'));
@@ -294,7 +294,7 @@ class TeamSpeak
                 }
 
                 // Give them a grace period
-                if(!$hasGracePeriod){
+                if (!$hasGracePeriod) {
                     self::pokeClient($client, trans('teamspeak.nickname.partiallyinvalid.poke1'));
                     self::pokeClient($client, trans('teamspeak.nickname.partiallyinvalid.poke2'));
                     self::messageClient($client, trans('teamspeak.nickname.partiallyinvalid.note', ['example' => $member->real_name.' - EGLL_N_TWR']));
--- a/app/Libraries/TeamSpeak.php
+++ b/app/Libraries/TeamSpeak.php
@@ -290,7 +290,6 @@ class TeamSpeak
                     Cache::forget(self::CACHE_NICKNAME_PARTIALLY_CORRECT.$client['client_database_id']);
                     Cache::forget(self::CACHE_NICKNAME_PARTIALLY_CORRECT_GRACE.$client['client_database_id']);
                     throw new ClientKickedFromServerException;
-                    return;
                 }
 
                 // Give them a grace period
--- a/app/Libraries/TeamSpeak.php
+++ b/app/Libraries/TeamSpeak.php
@@ -286,7 +286,7 @@ class TeamSpeak
                 Cache::forget(self::CACHE_NICKNAME_PARTIALLY_CORRECT.$client['client_database_id']);
                 Cache::forget(self::CACHE_NICKNAME_PARTIALLY_CORRECT_GRACE.$client['client_database_id']);
                 throw new ClientKickedFromServerException;
-            } elseif (!$hasGracePeriod) {
+            } elseif (!$hasGracePeriod && !$member->isDuplicateDisplayName($client['client_nickname'])) {
                 // set grace period to allow for possible mistakes
                 self::pokeClient($client, trans('teamspeak.nickname.partiallyinvalid.poke1'));
                 self::pokeClient($client, trans('teamspeak.nickname.partiallyinvalid.poke2'));
--- a/app/Libraries/TeamSpeak.php
+++ b/app/Libraries/TeamSpeak.php
@@ -275,26 +275,35 @@ class TeamSpeak
      */
     public static function checkClientNickname(TeamSpeak3_Node_Client $client, Account $member)
     {
-        if (!$member->isValidDisplayName($client['client_nickname'])) {
+        if (!$member->isValidDisplayName($client['client_nickname']) && !$member->isDuplicateDisplayName($client['client_nickname'])) {
             $recentlyTold = Cache::has(self::CACHE_NICKNAME_PARTIALLY_CORRECT.$client['client_database_id']);
             $hasGracePeriod = Cache::has(self::CACHE_NICKNAME_PARTIALLY_CORRECT_GRACE.$client['client_database_id']);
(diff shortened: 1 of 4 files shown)


app/Models/Mship/Account.php
```diff
@@ -437,31 +457,25 @@ public function getFullyDefinedAttribute()
...
     public function isPartiallyValidDisplayName($displayName)
     {
-        $allowedNames = collect();
-        $allowedNames->push($this->name);
-        $allowedNames->push($this->real_name);
-
-        return $allowedNames->filter(function ($item, $key) use ($displayName) {
+        return $this->allowedNames()->filter(function ($item, $key) use ($displayName) {
             return strstr(strtolower($displayName), strtolower($item)) != false;
         })->count() > 0;
```
AxonC:
```suggestion
        })->isEmpty();
```

app/Models/Mship/Account.php
```diff
@@ -437,31 +457,25 @@ public function getFullyDefinedAttribute()
...
     }
 
+    public function isDuplicateDisplayName($displayName)
+    {
+        return $this->allowedNames(true, true)->filter(function ($item, $key) use ($displayName) {
+            return preg_match("/^".$item."$/i", $displayName) == 1;
+        })->count() > 0;
```
AxonC:
```suggestion
        })->isEmpty();
```

app/Libraries/TeamSpeak.php
```diff
@@ -275,26 +275,35 @@ public static function checkMemberMandatoryNotifications(TeamSpeak3_Node_Client
...
+                }
+
+                // Give them a grace period
+                if (!$hasGracePeriod) {
+                    self::pokeClient($client, trans('teamspeak.nickname.partiallyinvalid.poke1'));
+                    self::pokeClient($client, trans('teamspeak.nickname.partiallyinvalid.poke2'));
+                    self::messageClient($client, trans('teamspeak.nickname.partiallyinvalid.note', ['example' => $member->real_name.' - EGLL_N_TWR']));
+
+                    $now = Carbon::now();
+                    Cache::put(self::CACHE_NICKNAME_PARTIALLY_CORRECT.$client['client_database_id'], $now, 6 * 60);
```
AxonC: I'd be tempted to write some form of cache wrapper here (and elsewhere). I never like to see a direct call to the cache. Will leave this up to you though as I feel it might be a sizable job.

app/Libraries/TeamSpeak.php
```diff
@@ -275,26 +275,35 @@ public static function checkMemberMandatoryNotifications(TeamSpeak3_Node_Client
      */
     public static function checkClientNickname(TeamSpeak3_Node_Client $client, Account $member)
     {
-        if (!$member->isValidDisplayName($client['client_nickname'])) {
+        if (!$member->isValidDisplayName($client['client_nickname']) && !$member->isDuplicateDisplayName($client['client_nickname'])) {
```
AxonC: Can you drop this onto two lines to make it a bit more readable?

app/Libraries/TeamSpeak.php
```diff
@@ -275,26 +275,35 @@ public static function checkMemberMandatoryNotifications(TeamSpeak3_Node_Client
...
+
+                // Check if they have been recently told but the grace has elapsed
+                if ($recentlyTold && !$hasGracePeriod) {
+                    self::pokeClient($client, trans('teamspeak.nickname.invalid.poke1'));
+                    self::pokeClient($client, trans('teamspeak.nickname.invalid.poke2'));
+                    self::kickClient($client, trans('teamspeak.nickname.invalid.kick'));
+                    Cache::forget(self::CACHE_NICKNAME_PARTIALLY_CORRECT.$client['client_database_id']);
+                    Cache::forget(self::CACHE_NICKNAME_PARTIALLY_CORRECT_GRACE.$client['client_database_id']);
+                    throw new ClientKickedFromServerException;
+                    return;
```
AxonC: Is the return needed when you are throwing an exception anyway?

app/Models/Mship/Account.php
```diff
@@ -437,31 +457,25 @@ public function getFullyDefinedAttribute()
...
-
-        if ($this->networkDataAtcCurrent) {
-            $allowedNames->push($this->name.' - '.$this->networkDataAtcCurrent->callsign);
-            $allowedNames->push($this->real_name.' - '.$this->networkDataAtcCurrent->callsign);
-        }
-
-        return $allowedNames->filter(function ($item, $key) use ($displayName) {
+        return $this->allowedNames(true)->filter(function ($item, $key) use ($displayName) {
             return strcasecmp($item, $displayName) == 0;
         })->count() > 0;
```
AxonC:
```suggestion
        })->isEmpty();
```

tests/Unit/Account/AccountModelTest.php
```diff
@@ -76,6 +76,14 @@ public function itCorrectlyDeterminesIfTheNicknameIsAValidDisplayName()
         $this->assertTrue($this->user->isValidDisplayName($fullNickname));
     }
 
+    /** @test */
+    public function itAllowsValidIncrementedNicknames()
+    {
+        $this->assertTrue($this->user->isDuplicateDisplayName($this->user->name_first . ' ' . $this->user->name_last . '1'));
```
AxonC: Maybe string interpolation here? "{$a} {$b}"

app/Models/Mship/Account.php
```diff
@@ -428,6 +428,26 @@ public function getFullyDefinedAttribute()
...
+        if ($withNumberWildcard) {
+            $wildcard = "\d";
+        }
+
+        $allowedNames = collect();
+        $allowedNames->push($this->name.$wildcard);
+        $allowedNames->push($this->real_name.$wildcard);
+
+        if ($includeATC && $this->networkDataAtcCurrent) {
+            $allowedNames->push($this->name.$wildcard.' - '.$this->networkDataAtcCurrent->callsign);
```
AxonC: String interpolation?

tests/Unit/TeamSpeakTest.php
```diff
@@ -100,30 +100,32 @@ public function testGroupQualification()
...
 
     public function testNotPartiallyValidDisplayName()
     {
-        $account = $this->account;
         $validDisplayName = 'John Do - EGKK_TWR';
-        $this->assertEquals(false, $account->isPartiallyValidDisplayName($validDisplayName));
+        $this->assertFalse($this->account->isPartiallyValidDisplayName($validDisplayName));
+    }
+
+    public function testAllowedDuplicates()
```
AxonC: How does this test allowed duplicates? As far as I can see, these would be allowed anyway.

app/Models/Mship/Account.php
```diff
@@ -437,31 +457,25 @@ public function getFullyDefinedAttribute()
...
 
     public function isPartiallyValidDisplayName($displayName)
     {
-        $allowedNames = collect();
-        $allowedNames->push($this->name);
-        $allowedNames->push($this->real_name);
-
-        return $allowedNames->filter(function ($item, $key) use ($displayName) {
+        return $this->allowedNames()->filter(function ($item, $key) use ($displayName) {
             return strstr(strtolower($displayName), strtolower($item)) != false;
```
AxonC: So basically these would be true? (!= false) 😉

tests/Unit/TeamSpeakTest.php
atoff: `John Do` vs `John Doe`. Checking that it doesn't except it just because there is a number at the end

app/Libraries/TeamSpeak.php
atoff: I personally don't think its too long / complicated, and feel if it was split into two lines it would just leave redundant braces and clutter it up? Its probably on the shorted end of some of our if's :P

app/Libraries/TeamSpeak.php
atoff: This is probably outside of the scope - I am sure we can come back to this later (I do not have the will power to do more TeamSpeak messing right now!)

app/Models/Mship/Account.php
atoff: I'm note sure I understand what you mean?

AxonC: Evaluating != false is the same as something being true, but then I forgot strstr() returns false if nothing is found. Gotta love inconsistent return types.

PR #1485, opened 2019-07-24
CORE-405 / CORE-417 - FTE Image Uploading Bug Fixes

atoff:
Note, will require existing images to be move from public/storage/smartcars... to public/smartcars... 

Closes CORE-405 and CORE-417

### DIFF
--- a/app/Libraries/Storage/CoreUploadedFile.php
+++ b/app/Libraries/Storage/CoreUploadedFile.php
@@ -16,7 +16,7 @@ class CoreUploadedFile extends UploadedFile
         return sha1("{$this->getClientOriginalName()}.{$this->getClientOriginalExtension()}");
     }
 
-    public function getFilePathName()
+    public function getPathFileName()
     {
         return $this->getFullFileName() . '.' . $this->extension();
     }
--- a/app/Libraries/Storage/CoreUploadedFile.php
+++ b/app/Libraries/Storage/CoreUploadedFile.php
@@ -13,11 +13,11 @@ class CoreUploadedFile extends UploadedFile
 
     public function getFullFileName()
     {
-        return sha1("{$this->getClientOriginalName()}");
+        return sha1("{$this->getClientOriginalName()}.{$this->getClientOriginalExtension()}");
     }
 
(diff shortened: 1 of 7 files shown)


AxonC:
Tests failing old chap @atoff

app/Libraries/Storage/CoreUploadedFile.php
```diff
@@ -15,4 +15,9 @@ public function getFullFileName()
     {
         return sha1("{$this->getClientOriginalName()}.{$this->getClientOriginalExtension()}");
     }
+
+    public function getFilePathName()
+    {
+        return $this->getFullFileName() . '.' . $this->extension();
```
AxonC: Consistency with the method above with respect to string construction please me duck

app/Libraries/Storage/FteStorageWrapper.php
```diff
@@ -4,7 +4,7 @@
 
 class FteStorageWrapper extends StorageWrapper
 {
-    protected $basePath = "storage/smartcars/exercises";
+    protected $basePath = "smartcars/exercises";
```
AxonC: Why are we changing the path?

atoff: Because, to me at least, it doesn't make sense to store in public/storage/storage/smartcars...

AxonC: I think that was done due to some of the symlinking in prod; where storage/storage corresponded to public/storage

atoff:
> Tests failing old chap @atoff

that would be because I forgot to commit the test update :///

Should be good now

app/Libraries/Storage/FteStorageWrapper.php
atoff: it should be that storage/public maps to public/storage. If its not this is a prod issue not standard practice, which we probably need to look at fixing

atoff:
@AxonC I've merged changes for CORE-417 in the latest commit (a8700ad) - just fixes the deleting aspect aswell

PR #1489, opened 2019-08-03
CORE-419 IPB Club Syncing

atoff:
Closes CORE-419


atoff:
@CalumTowers are you somehow able to test this to: 1) ensure it doesn't break ;) and 2) it does fix the original issue?

I don't know why the original error was occurring, because IPB should be able to find the club that it gave us in the first place (!) but it must have something to do with the new club.

PR #1490, opened 2019-08-03
Improvement: Send query bindings to bugsnag

atoff:
Allows the bindings for queries to be sent to bugsnag for faster debugging


CalumTowers:
I'll add this to the production env.

PR #1492, opened 2019-08-11
CORE-401 - Establish Waiting List Permissions

AxonC:


database/migrations/2019_07_20_085104_add_waiting_list_table.php
```diff
@@ -89,6 +89,17 @@ public function up()
             'created_at' => \Carbon\Carbon::now(),
             'updated_at' => \Carbon\Carbon::now(),
         ]);
+
+        $this->createPermission('waitingLists/*');
+        $this->createPermission('waitingLists/base');
+        $this->createPermission('waitingLists/atc/base');
```
CalumTowers: Do we need to be a little more descriptive on naming here? What is base actually doing? Can we just have ATC lists, Pilot lists and then “elevated” for special functions? Keen to keep as simple as possible.

AxonC: These were directly from the requirements from DC. You can see how the wildcards are used within the policy file.

CalumTowers: Yeah I've read the policy but still not convinced the "base" permission makes a tonne of sense. The users of the permission system should not be expected to have read our policies to understand what each permission is doing, nor read a spec from its formation. Can we review the naming conventions before merging please.

AxonC: Would you be happy with just having the departmental base permissions?

CalumTowers: Yeah that makes sense then maybe another permission for elevated stuff?

AxonC: Changes made @CalumTowers.

PR #1548, opened 2019-10-16
CORE-429 - Sync TG Members To Forum Groups

CalumTowers:


app/Jobs/TGForumGroups/SyncToForumGroup.php
```diff
@@ -0,0 +1,62 @@
...
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $ipboard = new Ipboard();
+
+        // Get the IP Board id based on the CID provided
+        // Need to think of a clever and clean way to get this...
```
CalumTowers: @atoff Any advice on this? I know you've done a fair amount of work on the Forum sync stuff in the past... I need to get the IPB member id based on the CID that is passed in?

CalumTowers: Would something like this work to load the Member object itself?

```
$member = \IPS\Member::load( $this->cid, 'p.field_12' );
```

Ref: https://invisioncommunity.com/4docs/advanced-usage/development/working-with-members-r179/

PR #1647, opened 2020-01-09
Updates message when kicking an inactive member from Teamspeak

cpawley:
image

When Kicking an inactive member, the message is truncated due to Teamspeak message length limits. This is a shorter message.


AxonC:
80 character limit confirmed here: https://forum.teamspeak.com/threads/78118-Max-channel-name-description-message-poke-length?p=328937#post328937

Thanks, @cpawley!

PR #1655, opened 2020-01-12
Bug Fix: TeamSpeak kicking without grace period

### LINKED ISSUE #1654: Bug - TeamSpeak nickname cooldowns
When a client doesn't have a grace period or 'recently told' cache entry, one is created if they meet the criteria. However, because the check for these cache entries is before this in the code, the software will kick instantly.

Tldr: set recently told and grace period booleans to true here: https://github.com/VATSIM-UK/core/blob/master/app/Libraries/TeamSpeak.php#L289

atoff:
Closes #1654


app/Libraries/TeamSpeak.php
```diff
@@ -288,6 +288,8 @@ public static function checkClientNickname(TeamSpeak3_Node_Client $client, Accou
                     $now = Carbon::now();
                     Cache::put(self::CACHE_NICKNAME_PARTIALLY_CORRECT.$client['client_database_id'], $now, 6 * 60);
                     Cache::put(self::CACHE_NICKNAME_PARTIALLY_CORRECT_GRACE.$client['client_database_id'], $now, 3 * 60);
+                    
+                    $recentlyTold = $hasGracePeriod = true;
```
AxonC: Why the second assignment? You appear to be setting the `recentlyTold` variable to the value of `hasGracePeriod`, and then set that to true. This will set both to be true, but sorry... I have no idea what this achieves at first glance, :(

atoff: Briefly explained in the issue - https://github.com/VATSIM-UK/core/issues/1654

If you expand the code, these two variables are dependent on the cache having the items that are set on lines 289-290. Logic after this `if` statement depends on these variables in deciding if the user has a kicking grace period or not. As these cache items have just been set, we should set these variables both to be true. This has been causing issues where members are assigned their grace period, but instantly kicked (ref. helpdesk ticket)

AxonC: Could you not just step out of the whole function at this point though? If you are setting a grace period, do you need to inform them of this, since they will still have had the first notification? The comment below even says that the notification occurd when the grace period is over - why continue the check if it isn't over? 

At least then you can re-trigger the code when the cache won't have an effective race condition. I might be completely wrong here, but this is my thoughts after looking at it for a quite a bit now.

AxonC: In the absence of a better solution as discussed, I'll approve this, but I'd like a second opinion.

PR #1673, opened 2020-01-30
cache the result of the hour check

AxonC:
Requires Redis config on core

### DIFF
--- a/app/Models/Training/WaitingList/WaitingListAccount.php
+++ b/app/Models/Training/WaitingList/WaitingListAccount.php
@@ -8,7 +8,7 @@ use App\Models\Training\WaitingList;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Relations\Pivot;
 use Illuminate\Database\Eloquent\SoftDeletes;
-use Illuminate\Support\Facades\Redis;
+use Illuminate\Support\Facades\Cache;
 
 class WaitingListAccount extends Pivot
 {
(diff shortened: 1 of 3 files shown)


AxonC:
> Whilst I appreciate the merits of Redis for our production server, can you clarify why it is required for this change?
> From what I can see we could just use the Laravel Cache driver to store this to our current cache system and then change it at a later date to Redis for the entire app?

The main motivation is that much of nova leverages Redis anyway, so taking the plunge in small doses might be more effective. I can change it to just using the cache driver if you like, but this is just a database table which I don't see actually providing much of a benefit over what is already occurring. Compare that to Redis which is a dedicated store for this data.

CalumTowers:
> > Whilst I appreciate the merits of Redis for our production server, can you clarify why it is required for this change?
> > From what I can see we could just use the Laravel Cache driver to store this to our current cache system and then change it at a later date to Redis for the entire app?
> 
> The main motivation is that much of nova leverages Redis anyway, so taking the plunge in small doses might be more effective. I can change it to just using the cache driver if you like, but this is just a database table which I don't see actually providing much of a benefit over what is already occurring. Compare that to Redis which is a dedicated store for this data.

I’d like to switch to the Cache driver so any changes we do to migrate to Redis are done in one place (e.g. config). Cache in the database would still be quicker than querying the network data tables for every person and every position as the cache would just return true or false for each flag.

AxonC:
> > > Whilst I appreciate the merits of Redis for our production server, can you clarify why it is required for this change?
> > > From what I can see we could just use the Laravel Cache driver to store this to our current cache system and then change it at a later date to Redis for the entire app?
> > 
> > 
> > The main motivation is that much of nova leverages Redis anyway, so taking the plunge in small doses might be more effective. I can change it to just using the cache driver if you like, but this is just a database table which I don't see actually providing much of a benefit over what is already occurring. Compare that to Redis which is a dedicated store for this data.
> 
> I’d like to switch to the Cache driver so any changes we do to migrate to Redis are done in one place (e.g. config). Cache in the database would still be quicker than querying the network data tables for every person and every position as the cache would just return true or false for each flag.

Changed to the `Cache` facade

PR #1677, opened 2020-02-01
Add / enable feedback to new admin panel

AxonC:
Adds a feedback module to Nova

Can't prefil actions against feedback so added copy paste function

Also fixed bug I was getting on PHP 7.4 in the feedback controller


AxonC:
@CalumTowers opened PR for review, but still worth testing locally. I've done all the perms, but you should be able to check this out and it just work

CalumTowers:
> @CalumTowers opened PR for review, but still worth testing locally. I've done all the perms, but you should be able to check this out and it just work

Back over to you following my amends - let me know your thoughts.
I'm comfortable with us removing the functionality to configure forms from within the admin panel.

composer.json
```diff
@@ -100,6 +100,7 @@
         "planetteamspeak/ts3-php-framework": "1.1.33",
         "predis/predis": "^1.1",
         "pusher/pusher-php-server": "~4.0",
+        "sixlive/nova-text-copy-field": "^1.5",
```
AxonC: Will need moving in future

AxonC:
> 
> 
> > @CalumTowers opened PR for review, but still worth testing locally. I've done all the perms, but you should be able to check this out and it just work
> 
> Back over to you following my amends - let me know your thoughts.
> I'm comfortable with us removing the functionality to configure forms from within the admin panel.

Only thing I can see missing is permission to view on feedback (or not as it might be)

composer.json
CalumTowers: We can move as part of the other PR yep

CalumTowers:
> > > @CalumTowers opened PR for review, but still worth testing locally. I've done all the perms, but you should be able to check this out and it just work
> > 
> > 
> > Back over to you following my amends - let me know your thoughts.
> > I'm comfortable with us removing the functionality to configure forms from within the admin panel.
> 
> Only thing I can see missing is permission to view on feedback (or not as it might be)

The perms appeared to work fine when I tested locally for viewing the different forms. Can you double check?

AxonC:
> > > > @CalumTowers opened PR for review, but still worth testing locally. I've done all the perms, but you should be able to check this out and it just work
> > > 
> > > 
> > > Back over to you following my amends - let me know your thoughts.
> > > I'm comfortable with us removing the functionality to configure forms from within the admin panel.
> > 
> > 
> > Only thing I can see missing is permission to view on feedback (or not as it might be)
> 
> The perms appeared to work fine when I tested locally for viewing the different forms. Can you double check?

Sorry, stupid typo. It meant to say 'own' feedback.

app/Policies/Nova/FeedbackPolicy.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        if (parent::before($account, $policy)) {
+            return true;
+        }
+    }
+
+    public function view(Account $account, Feedback $feedback)
+    {
+        return $account->checkPermissionTo("feedback/view/{$feedback->form->slug}", self::GUARD)
+            && $feedback->account_id != $account->hiddenFeedbackUsers();
```
CalumTowers: Former is an int, latter is an array. Comparison will fail.

PR #1680, opened 2020-02-01
Nova Without Nova

CalumTowers:
To deploy Core without Nova (e.g. for those without the auth details for Nova) you can now;

- Delete line 142 from `composer.json` which will stop Nova specific packages (defined in `composer-nova.json`)
- Run `composer update` and `composer install` to regenerate `composer.lock` and install dependencies locally.

Caution should be taken that the revised `composer.lock` from this is not deployed to `master`.

Off the back of some recent difficulties with deployment locally, various amends to boilerplate files and the readme.


AxonC:
Worth adding a note into the readme on how to setup without Nova access.

CalumTowers:
> Worth adding a note into the readme on how to setup without Nova access.

Yep but we could do with writing a proper deployment guide (with and without) - will add that to the backlog.

Issue #1562, opened 2019-10-18
As a user, I only need one UKCP API key


CalumTowers:
Look at allowing users just 1 key, and prompt to redownload / delete key when requesting a new one.

Linked UKCP API issue: https://github.com/VATSIM-UK/uk-controller-api/issues/41

PR #1681, opened 2020-02-02
Add guides in to dashboard

aaronholtomcook:
Small change to add the guide URLs in the panel footer. 
Let me know what you guys think, or if you want a separate guides page setting up with the content copied over I can do that.


AxonC:
Closes #1562.

PR #1682, opened 2020-02-02
Start Decommissioning Admin Panel

CalumTowers:


app/Http/Controllers/Adm/Dashboard.php
```diff
@@ -76,7 +30,7 @@ public function anySearch($searchQuery = null)
         }
 
         if (strlen($searchQuery) < 2 or $searchQuery == null) {
```
aaronholtomcook: `strlen` should accept the null value, so you can just write `strlen($searchquery) < 2`

app/Nova/Feedback.php
```diff
@@ -55,23 +65,26 @@ class Feedback extends Resource
...
             Text::make('Feedback Form', function () {
                 return $this->form->name;
             }),
 
+            BelongsTo::make('Subject', 'account', 'App\Nova\Account'),
+
+            BelongsTo::make('Submitted By', 'submitter', 'App\Nova\Account')
+                ->canSeeWhen("seeSubmitter", $this),
+
+            DateTime::make('Submitted At', 'created_at')->onlyOnDetail()->format('Do MMMM YYYY HH:mm'),
```
aaronholtomcook: Would be a good idea to move the date formatting string to a static class member to use elsewhere.

`
public static class DateFormat
{
    public const string DefaultFormat = "Do MMMM YYYY HH:mm";
}
`

`
DateTime::make('Submitted At', 'created_at')->onlyOnDetail()->format(DateFormat.DefaultFormat),
`

CalumTowers: We're driving this with Laravel Nova so it's a little limited in that regard. My view is it would be better to continue using the Nova supported `->format` method, but interested to hear @AxonC's view!

app/Http/Controllers/Adm/Dashboard.php
CalumTowers: This PR is the start of us decommissioning the admin panel so this code will soon be deleted and replaced with Nova 😄

app/Nova/Feedback.php
AxonC: I'm easy on this one - we are constrained by the software here, but defining it as a const within that resource, or a base resource wouldn't be a bad thing.

PR #1683, opened 2020-02-02
Redirect user back to home page after 5 seconds.

aaronholtomcook:


CalumTowers:
@aaronholtomcook Thanks! 👍

PR #1706, opened 2020-02-28
Limit UKCP Tokens to 1

### LINKED ISSUE #1562: As a user, I only need one UKCP API key

atoff:
Refactors functionality to allow users only one API token.

As a note, instead of enforcing a hard limit, I have maintained the HasMany relationship style, so that people with existing multiple tokens aren't suddenly effected (and in case we ever want to do some with multiple tokens again). Over time, people with multiple tokens should be phased out naturally.

image

Closes #1562


app/Http/Controllers/UKCP/Token.php
```diff
@@ -35,47 +34,31 @@ public function create()
             return Redirect::route('mship.manage.dashboard')
                 ->withError('An unknown error occured, please contact Web Services.');
         }
-
-        $latestId = $this->ukcp->getValidTokensFor(auth()->user())->first()->id;
-        $tokenPath = 'ukcp/tokens/' . auth()->user()->id . '/' . $latestId . '.json';
-        Storage::disk('local')->put($tokenPath, $newToken);
-
-        return $this->viewMake('ukcp.token.create')->with('newToken', $latestId);
+        return redirect()->route('ukcp.guide')->withSuccess('Tokens Updated!');
```
AxonC: Do you really want to go back to the guide?

atoff: I think this is best. Provides a clear way to download the new key (the big download button), and for new people they don't need to click the link to view what they need to do (= hopefully less questions!)

atoff: And it already did go back to the guide (ukcp.token.create is the view that is now renamed ukcp.token.guide, and the ukcp.guide route makes that view)

app/Libraries/UKCP.php
```diff
@@ -3,31 +3,44 @@
...
 use App\Models\Mship\Account;
+use Bugsnag\BugsnagLaravel\Facades\Bugsnag;
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\ClientException;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Storage;
+use Psr\Http\Message\ResponseInterface;
 
 class UKCP
```
AxonC: Might be worth creating a singleton in the AppServiceProvider to auto-resolve the dependency injection

atoff: Did this originally, it doesn't make a difference. The dependency is automatically resolved anyway

AxonC: that would be the major motivation behind it

PR #1728, opened 2020-03-29
Fixing Waiting List Buckets

CalumTowers:


nova-components/WaitingListsManager/resources/js/components/Bucket.vue
```diff
@@ -29,7 +29,7 @@
                         </div>
                     </td>
                     <td>{{ account.id }}</td>
-                    <td>{{ this.moment(account.created_at).format("MMMM Do YYYY") }}</td>
+                    <td>{{ account.created_at }}</td>
```
CalumTowers:
```suggestion
                    <td>{{ this.moment(account.created_at).format("MMMM Do YYYY") }}</td>
```

Changing to this errors, reporting `moment` as undefined. I cannot see anything in the Nova docs that would cause this.
@AxonC Are you able to take a look? The latest commit on this branch means that the buckets will now load, with the only issue being that the `created_at` timestamp is displayed in its raw ISO format.

AxonC: should be sorted now

PR #1743, opened 2020-04-18
1708 - Implement Connect Login

### LINKED ISSUE #1708: New SSO
https://auth.vatsim.net/login

CalumTowers:
Resolves #1708


app/Http/Controllers/Auth/LoginController.php
```diff
@@ -36,161 +28,78 @@ class LoginController extends BaseController
...
+            return redirect()->route('dashboard')->withError('Something went wrong, please try again.');
         }
-        // Check if there was a CURL error code
-        if (VATSIMSSO::error()['code']) {
-            Log::error('VATSIMSSO was unable to reach CERT. Code:'.VATSIMSSO::error()['code'].' Message:'.VATSIMSSO::error()['message']);
 
-            return redirect()->back()->withErrors(['connection' => "We were unable to contact VATSIM's certification service. Please try again later. If this persists, please contact Web Services."]);
+        $resourceOwner = json_decode(json_encode($this->provider->getResourceOwner($accessToken)->toArray()));
+
+        if (
```
AxonC, line 62: Might be worth destructuring the parameters a little to prevent repeat references
```php
[$cid, $personal, $vatsim, $oauth] = $resourceOwner->data
```

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -36,161 +28,78 @@ class LoginController extends BaseController
...
+            ! $resourceOwner->data ||
+            ! $resourceOwner->data->cid ||
+            ! $resourceOwner->data->personal ||
+            ! $resourceOwner->data->personal->name_first ||
+            ! $resourceOwner->data->personal->name_last ||
+            ! $resourceOwner->data->personal->email ||
+            ! $resourceOwner->data ||
```
AxonC, line 69: This appears to be a duplicate of the one on line 63

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -36,161 +28,78 @@ class LoginController extends BaseController
...
-
-    public function vSsoValidationSuccess($user, $request)
-    {
-        $account = Account::firstOrNew(['id' => $user->id]);
-        $account->name_first = utf8_decode($user->name_first);
-        $account->name_last = utf8_decode($user->name_last);
-        $account->email = $user->email;
-        $account->experience = $user->experience;
-        $account->joined_at = $user->reg_date;
+        $account = Account::firstOrNew(['id' => $resourceOwner->data->cid]);
```
AxonC, line 85:
```php
[$name_first, $name_last, $email] = $resourceOwner->data;
```

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -36,161 +28,78 @@ class LoginController extends BaseController
...
+        Auth::guard('vatsim-sso')->loginUsingId($account->id);
 
-        Auth::login(Auth::guard('vatsim-sso')->user(), true);
-
-        return redirect($intended);
+        return SecondaryLoginController::attemptSecondaryAuth($account);
     }
 
-    public function loginSecondary(Request $request)
+    protected function completeLogin($resourceOwner, $token)
```
AxonC, line 83: Can we be any more specific about the types of these two parameters?

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -36,161 +28,78 @@ class LoginController extends BaseController
...
     {
-        if (Auth::guard('vatsim-sso')->check() && ! Auth::check()) {
-            return $this->attemptSecondaryAuth();
-        } else {
-            return redirect()->route('dashboard');
-        }
-    }
+        if (! $request->has('code') || ! $request->has('state')) {
+            $authorizationUrl = $this->provider->getAuthorizationUrl();
+            $request->session()->put('vatsimauthstate', $this->provider->getState());
```
AxonC, line 38: Worth putting `vatsimauthstate` into either a controller class constant or the config?

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -36,161 +28,78 @@ class LoginController extends BaseController
      */
     public function __construct()
     {
-        $this->middleware('guest')->except('showLoginForm', 'loginSecondary', 'logout');
+        $this->provider = new VatsimOAuthController();
```
AxonC, line 31: Because this is a controller, you should be able to inject this class as constructor param and have the service container resolve it for you. Having said that, you might need to bind a singleton. Give it a go!

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -36,161 +28,78 @@ class LoginController extends BaseController
...
-        }
+        $account = $this->completeLogin($resourceOwner, $accessToken);
 
-        $intended = Session::pull('url.intended', route('site.home'));
+        Auth::guard('vatsim-sso')->loginUsingId($account->id);
 
-        Auth::login(Auth::guard('vatsim-sso')->user(), true);
-
-        return redirect($intended);
+        return SecondaryLoginController::attemptSecondaryAuth($account);
```
AxonC, line 80: Would it not be more prudent to redirect them to the route that this controller method corresponds to? To me, this doesn't feel like the best approach redirecting to another class which has the sole purpose of handling HTTP requests. SRP and all.

app/Http/Controllers/Auth/LogoutController.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace App\Http\Controllers\Auth;
+
+use App\Http\Controllers\BaseController;
+use Illuminate\Support\Facades\Auth;
+
+class LogoutController extends BaseController
```
AxonC, line 8: Can this be an invokable, single method controller?

app/Http/Controllers/Auth/SecondaryLoginController.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Validate the user login request.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @return void
+     */
+    protected function validateLogin(Request $request)
+    {
```
AxonC, line 48: I can't see this being used anywhere - is it just providing some custom logic for the baseclass?

app/Http/Controllers/Auth/LoginController.php
AxonC: Further to this, the model binding should keep the account model in tact if it is in the correct state. Food for thought.

app/Http/Controllers/Auth/VATSIMOAuthController.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace App\Http\Controllers\Auth;
+
+use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
+use League\OAuth2\Client\Provider\GenericProvider;
+
+class VATSIMOAuthController extends GenericProvider
```
AxonC, line 8: Is this a controller if it is not serving any HTTP requests?

app/Http/Controllers/Auth/VATSIMOAuthController.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Initializes the provider variable.
+     */
+    public function __construct()
+    {
+        parent::__construct([
+            'clientId'                => config('vatsim-connect.id'),
+            'clientSecret'            => config('vatsim-connect.secret'),
+            'redirectUri'             => route('login.post'),
+            'urlAuthorize'            => config('vatsim-connect.base').'/oauth/authorize',
```
AxonC, line 24:
```suggestion
            'urlAuthorize'            => "{config('vatsim-connect.base')}'/oauth/authorize",
```
See if that works? Looks a little bit cleaner.

app/Http/Controllers/Auth/VATSIMOAuthController.php
```diff
@@ -0,0 +1,49 @@
...
+            'urlAccessToken'          => config('vatsim-connect.base').'/oauth/token',
+            'urlResourceOwnerDetails' => config('vatsim-connect.base').'/api/user',
+            'scopes'                  => config('vatsim-connect.scopes'),
+            'scopeSeparator'          => ' ',
```
AxonC, line 28: Just checking this is intended to be an empty string with a bit of whitespace (length of string is still technically zero since the `\0` is not counted.

app/Http/Controllers/Auth/VATSIMOAuthController.php
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public static function updateToken($token)
+    {
+        $controller = new self;
+        try {
+            return $controller->getAccessToken('refresh_token', [
+                'refresh_token' => $token->getRefreshToken(),
+            ]);
+        } catch (IdentityProviderException $e) {
+            return;
```
AxonC, line 46: Is there a better course of action here rather than just exiting the function? Or, should this exception be caught by consumers of the function?

composer.json
```diff
@@ -97,7 +94,7 @@
         "spatie/laravel-cookie-consent": "^2.4",
         "spatie/laravel-permission": "^3.0",
         "symfony/translation": "^5.0",
-        "vatsim/sso": "dev-develop#7579e9f492d84f04bfd30df41f0073b3d7a01527",
+        "vatsim/sso": "3.*",
```
AxonC, line 97: Can we be any more specific, maybe to a minor version at least? I know there shouldn't be any breaking changes, but that makes the dangerous assumption of SEMVER being followed to the spec.

app/Models/Mship/Account.php
```diff
@@ -512,4 +514,35 @@ public function __toString()
...
+                'refresh_token' => $this->vatsim_refresh_token,
+                'expires'       => $this->vatsim_token_expires,
+            ]);
+
+            if ($token->hasExpired()) {
+                $token = VatsimOAuthController::updateToken($token);
+            }
+
+            $this->update([
+                'vatsim_access_token'  => ($token) ? $token->getToken() : null,
```
AxonC, line 540:
```suggestion
                'vatsim_access_token'  => $token->getToken() || null,
```
See if that yields the same results

app/Models/Mship/Account.php
```diff
@@ -512,4 +514,35 @@ public function __toString()
...
+
+    /**
+     * Return a valid access token for the user or return null if none.
+     *
+     * @return \League\OAuth2\Client\Token\AccessToken
+     * @return null
+     */
+    public function getTokenAttribute()
+    {
+        if ($this->vatsim_access_token === null) {
```
AxonC, line 526: Could you invert this loop so the bulk of the logic is in the if part and the else is the secondary condition? Also might be worth throwing an exception instead of returning null.

app/Models/Mship/Account.php
```diff
@@ -512,4 +514,35 @@ public function __toString()
     {
         return $this->name;
     }
+
+    /**
+     * Return a valid access token for the user or return null if none.
+     *
+     * @return \League\OAuth2\Client\Token\AccessToken
+     * @return null
```
AxonC, line 522: Does it always return null?

AxonC:
There's a few things which sonar has flagged as well, I should probably up the quality gate to fail on any tbf: https://sonarcloud.io/project/issues?id=VATSIM-UK_core&pullRequest=1743&resolved=false&types=CODE_SMELL

app/Models/Mship/Account.php
CalumTowers: Line 521 gives you the other possible return type

AxonC: senior moment...

app/Http/Controllers/Auth/SecondaryLoginController.php
CalumTowers: This overrides a parent class so that the email field is not required on the secondary password

app/Http/Controllers/Auth/VATSIMOAuthController.php
CalumTowers: Not sure this is much cleaner, but it didn't work anyhow!

AxonC: might not even need the curly braces anymore, nae bother

app/Http/Controllers/Auth/VATSIMOAuthController.php
CalumTowers: I would presume that this is there to catch the exception that will be thrown if the user has revoked our access to their account on VATSIM's side.

app/Http/Controllers/Auth/VATSIMOAuthController.php
CalumTowers: This is from the example repo and separates the config out `'scopes' => explode(',', env('VATSIM_OAUTH_SCOPES')),`

app/Http/Controllers/Auth/LoginController.php
CalumTowers: We can type hint object, can't be much more specific than that without some additional complication

d590ec1b

app/Http/Controllers/Auth/LoginController.php
CalumTowers: 9cabd239

app/Http/Controllers/Auth/LogoutController.php
CalumTowers: 1dc053de

app/Http/Controllers/Auth/VATSIMOAuthController.php
CalumTowers: Nope! A provider...
3df10fb5

composer.json
CalumTowers: 21655fec

app/Models/Mship/Account.php
CalumTowers: Method removed in ecb238f8

app/Models/Mship/Account.php
CalumTowers: Removed the entire method as right now we don't need it - we only use SSO for login and not for updating user info. If we do in the future, we can re-add this logic.

Same with Discord. We store an access and refresh token but do not use them.

ecb238f8

CalumTowers:
> There's a few things which sonar has flagged as well, I should probably up the quality gate to fail on any tbf: https://sonarcloud.io/project/issues?id=VATSIM-UK_core&pullRequest=1743&resolved=false&types=CODE_SMELL

Two of the three things it flagged are Laravel-isms that are false positives.

`$provider` isn't used in the method but it does get initialised via the constructor and then used elsewhere
`config()` stuff being referenced three times is fine in my opinion?

app/Http/Controllers/Auth/LoginController.php
CalumTowers: Done!

AxonC:
> `config()` stuff being referenced three times is fine in my opinion?
Maybe worth just referencing it once to retrieve the value and then reference that variable?

app/Http/Controllers/Auth/LoginController.php
CalumTowers: I think this is fine as is, although the two suggestions would also be fine

app/Http/Controllers/Auth/LoginController.php
CalumTowers: Take a look at commit 938f9a7
It becomes pretty hard to test unless I extract this out to another method - thoughts?

CalumTowers:
> > `config()` stuff being referenced three times is fine in my opinion?
> 
> Maybe worth just referencing it once to retrieve the value and then reference that variable?

Not really, I'd be just extracting the `vatsim-connect.base` config option to solve the duplication. Then I'd need to extract the id, secret and scope for consistency. Feels cleaner to just reference the config option multiple times

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -22,89 +20,64 @@ class LoginController extends BaseController
...
-        // user has not been authenticated with VATSIM SSO
-        if (! Auth::guard('vatsim-sso')->check()) {
-            return $this->attemptVatsimAuth();
+            return redirect()->away($authorizationUrl);
         }
 
-        if (! Auth::check()) {
-            $this->attemptSecondaryAuth();
+        if ($request->input('state') !== session()->pull('vatsimauthstate')) {
+            return redirect()->route('dashboard')->withError('Something went wrong, please try again.');
```
atoff, line 45: The user won't have access to dashboard here (I think) - so this will just redirect somewhere else? For efficiency, we should redirect direct to a page they have access to

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -22,89 +20,64 @@ class LoginController extends BaseController
...
-        if (VATSIMSSO::error()['code']) {
-            Log::error('VATSIMSSO was unable to reach CERT. Code:'.VATSIMSSO::error()['code'].' Message:'.VATSIMSSO::error()['message']);
 
-            return redirect()->back()->withErrors(['connection' => "We were unable to contact VATSIM's certification service. Please try again later. If this persists, please contact Web Services."]);
+        $resourceOwner = json_decode(json_encode($this->provider->getResourceOwner($accessToken)->toArray()));
+
+        if (
+            ! $resourceOwner->data ||
+            ! $resourceOwner->data->cid ||
+            ! $resourceOwner->data->personal ||
```
atoff, line 66: Would recommend we use https://laravel.com/docs/7.x/helpers#method-data-get coupled with a "required" array to do this. Maybe a bit more neat (although a bit more complex)

e.g.
```
 $needed = collect([
            'cid',
            'personal.name_first',
            'personal.name_last',
             ...
        ]);

        $hasAllRequired = $needed->filter(function ($path) use ($resourceOwner){
            return data_get($resourceOwner->data, $path) == null;
        })->empty()

        if(!$hasAllRequired ){
            return $this->onNotEnoughDetailsError($request);
        }
```

app/Http/Controllers/Auth/LogoutController.php
```diff
@@ -0,0 +1,16 @@
...
+{
+    public function __invoke()
+    {
+        Auth::logout();
+
+        return redirect(route('site.home'));
```
atoff, line 14: For neatness :)
```suggestion
        return redirect()->route('site.home');
```

composer.json
```diff
@@ -75,7 +71,7 @@
         "cobaltgrid/vatsim-stand-status": "~1.0",
         "doctrine/dbal": "~2.10",
         "guzzlehttp/guzzle": "~6.0",
-        "laravel/framework": "^7.0",
+        "laravel/framework": "7.23.2",
```
atoff, line 74: This is restricting us to just this version? Wouldn't we want `^7.23` - not sure if there is a specific reason for locking to 7.23.2?

atoff (dismissed):

composer.json
CalumTowers: There is a bug in 7.24 re the Spatie Permission package. Its being fixed on another PR but I've fixed the version so I can run this branch locally

app/Http/Controllers/Auth/LoginController.php
CalumTowers: Tried playing around with this and seemed to cause a mild faff. Will stick with it as is.

app/Http/Controllers/Auth/LoginController.php
CalumTowers: NFA

app/Http/Controllers/Auth/LoginController.php
CalumTowers: Dashboard is available without auth, it is this page

image

PR #1745, opened 2020-04-19
Feature - Endorsement cache bust on ATC session end

atoff:
Busts a user's endorsement cache when a ATC session ends.

Currently, the overall status for an endorsement does not tally with the status for the individual conditions. This fix will cache-bust the user's endorsement-met status when their ATC session ends.

### DIFF
--- a/app/Listeners/NetworkData/FlushEndorsementCache.php
+++ b/app/Listeners/NetworkData/FlushEndorsementCache.php
@@ -13,8 +13,8 @@ class FlushEndorsementCache implements ShouldQueue
     public function handle(AtcSessionEnded $event)
     {
         $user = $event->atcSession->account;
-        Endorsement::pluck('id')->each(function ($id) use ($user) {
-            Cache::forget(Endorsement::generateCacheKey($id, $user->id));
+        Endorsement::get(['id'])->each(function (Endorsement $endorsement) use ($user) {
+            Cache::forget($endorsement->generateCacheKey($user));
         });
     }
 }
--- a/app/Listeners/NetworkData/FlushEndorsementCache.php
+++ b/app/Listeners/NetworkData/FlushEndorsementCache.php
@@ -13,7 +13,7 @@ class FlushEndorsementCache implements ShouldQueue
     public function handle(AtcSessionEnded $event)
     {
         $user = $event->atcSession->account;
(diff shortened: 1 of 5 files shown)


app/Models/Atc/Endorsement.php
```diff
@@ -38,4 +38,9 @@ public function conditionsMetForUser(Account $account): bool
 
         return $allMet;
     }
+
+    public static function generateCacheKey($endorsementId, $accountId)
```
AxonC, line 42: Strong types on the parameters?

app/Listeners/NetworkData/FlushEndorsementCache.php
```diff
@@ -0,0 +1,22 @@
...
+use App\Notifications\AtcSessionRecordedConfirmation;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Support\Facades\Cache;
+
+class FlushEndorsementCache implements ShouldQueue
+{
+    public function handle(AtcSessionEnded $event)
+    {
+        $user = $event->atcSession->account;
+        $endorsementIds = Endorsement::pluck('id');
```
AxonC, line 16: Can convert to each loop since pluck returns a collection from memory? Know it is a small thing but open to extension and all that

app/Models/Atc/Endorsement.php
CalumTowers: I'd agree and perhaps convert to `$endorsement` and `$account` to pass in the full object

app/Listeners/NetworkData/FlushEndorsementCache.php
atoff: Nope good idea

app/Models/Atc/Endorsement.php
atoff: > I'd agree and perhaps convert to `$endorsement` and `$account` to pass in the full object

If we did this, in the `FlushEndorsementCache` listener, we would have to do `Endorsement::get(['id'])->...` so that an instance of the Endorsement is created. Happy if I just leave it as passing in a full account model, but just the endorsement id?

CalumTowers: Sure that works

atoff: Think I have found a solution that is a better combination of both!

PR #1746, opened 2020-04-20
Update frequencies on operations.sectors

hsugden:
Fixed some outdated frequencies, added some new ones to the tables
Corrected top-down order


hsugden:
Tables before:

image

Tables after:

image

PR #1754, opened 2020-04-27
Downgrade Nova

CalumTowers:
- Fixed "Actions" appeariing
- Fixes issues with availableForNavigation always returning true


AxonC:
Is this a downgrade or upgrade? 😕

CalumTowers:
> Is this a downgrade or upgrade? 😕

Downgrade by one patch version as the .1 is buggy at the moment apparently!

PR #1763, opened 2020-05-07
Update oceanic procedures document link

hsugden:
Updated link


hsugden:
In fact, this is the wrong one! 😄 One moment...

PR #1764, opened 2020-05-07
Update oceanic procedures document link

hsugden:


CalumTowers:
Thanks Harry. We are in the process of reworking the Downloads section so links are likely to change. Our hope is this will make it more clear where to find things.

PR #1774, opened 2020-05-11
Add Redis Config Options

CalumTowers:
- Adds different options for Redis so that we can split the `session`, `cache` and `queue` into separate Redis databases
- Adds Laravel Horizon to monitor Redis queues
- Upgrades Nova to the latest version to ensure compatibility with Laravel Horizon's latest version


resources/views/vendor/nova/partials/meta.blade.php
```diff
@@ -1,7 +1 @@
 {{-- <link rel="icon" type="image/png" href="{{ asset('/img/favicon.png') }} "> --}}
-<style>
```
AxonC, line 2: Just for future reference, this is the fix for the feedback overflow.

PR #1784, opened 2020-05-21
Style: Reflect new staff changes

Keanu73:
Reflect new TG restructure & new vacancies


inventor02:
seeing as we are waiting on 2 new appointments doesn't it just make sense to wait until then?

Keanu73:
probably

PR #1794, opened 2020-05-27
Add Support For GitHub Actions On Forks

CalumTowers:
Commit messages should give a step-by-step of the changes made to make this work.

This PR is from my fork so should be sufficient as a proof of concept - it builds and passes without ever pulling Nova in or access to private tokens.

StyleCI can only run on our branches. If we merged something to master that did not comply with our rules, StyleCI would apply the fixed to master automatically.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -92,7 +92,6 @@
         "planetteamspeak/ts3-php-framework": "1.1.33",
         "predis/predis": "^1.1",
         "pusher/pusher-php-server": "~4.0",
-        "sixlive/nova-text-copy-field": "^1.5",
         "skymeyer/vatsimphp": "^2.0",
         "spatie/laravel-cookie-consent": "^2.4",
         "spatie/laravel-permission": "^3.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -85,14 +85,14 @@
...
         "laravel/tinker": "^2.0",
+        "laravel/ui": "^2.0",
         "laravelcollective/html": "^6.0",
         "league/csv": "^8.0",
         "maatwebsite/excel": "~3.1.17",
         "malahierba-lab/public-id": "dev-master",
         "planetteamspeak/ts3-php-framework": "1.1.33",
         "predis/predis": "^1.1",
         "pusher/pusher-php-server": "~4.0",
-        "sixlive/nova-text-copy-field": "^1.5",
```
AxonC, line 95: removing this will break the feedback resource

CalumTowers: It is in `composer-nova.json` which is merged to the main `composer.json` when required. It has been referenced in both files for some reason, this only removes the duplicate.

PR #1795, opened 2020-05-29
refactor: remove reference to env in teamspeak create view

AxonC:
Fixes 1776.

@CalumTowers I'm unable to test this locally since I don't have a valid TS instance to connect too (see screenshot) 

image

image


app/Http/Controllers/TeamSpeak/Registration.php
```diff
@@ -37,10 +37,13 @@ public function getNew()
         $autoURL = 'ts3server://'.config('services.teamspeak.host').'?nickname='.$this->account->name_first.'%20';
         $autoURL .= $this->account->name_last.'&token='.$confirmation->privilege_key;
 
+        $teamspeakURL = config('teamspeak.host');
```
CalumTowers, line 40: I know this controller method is already horrendous, but do you think this would look better inline?

`->with('teamspeak_url', config('teamspeak.host'))`

Just to try and avoid more temp variables?

AxonC: good call, changed

PR #1868, opened 2020-08-01
Fix Homepage Mobile Menu

### LINKED ISSUE #1864: Homepage Hamburger Menu
The hamburger menu on mobile devices does not appear to function when clicked.

CalumTowers:
Fixed #1864


AxonC:
What's changing here? It's showing an 11k+ changeset

CalumTowers:
> What's changing here? It's showing an 11k+ changeset

It reverts the changes that we did as part of removing Google Analytics. But this file didn’t need to change and the amend caused a break. The formatting of this file is also pretty weird so may be easier to pull locally and review that way.

Issue #1858, opened 2020-07-22
Handle Missing UKCP Token


vatsim-uk-pa:
## Error in Core

**League\Flysystem\FileNotFoundException** in **GET /ukcp/token/XXXXX/download**
File not found at path: ukcp/tokens/XXXXXXX/XXXXX

[View on Bugsnag](https://app.bugsnag.com/vatsim-uk/core/errors/5f1624fede15e000188b758e?event_id=5f1624fe005b890cb9da0000&i=gh&m=ci)

## Stacktrace

    app/Http/Controllers/UKCP/Token.php:60 - App\Http\Controllers\UKCP\Token::download
    app/Http/Controllers/BaseController.php:40 - App\Http\Controllers\BaseController::App\Http\Controllers\{closure}
    app/Http/Middleware/RedirectToIntended.php:23 - App\Http\Middleware\RedirectToIntended::handle
    app/Traits/Middleware/RedirectsOnFailure.php:48 - App\Http\Middleware\MustHaveCommunityGroup::handle
    app/Traits/Middleware/RedirectsOnFailure.php:48 - App\Http\Middleware\UserMustReadNotifications::handle
    app/Http/Middleware/DenyIfBanned.php:27 - App\Http\Middleware\DenyIfBanned::handle
    app/Traits/Middleware/RedirectsOnFailure.php:48 - App\Http\Middleware\MandatoryPasswords::handle
    app/Http/Middleware/RecordLoginInfo.php:39 - App\Http\Middleware\RecordLoginInfo::handle
    app/Http/Middleware/TrackInactivity.php:43 - App\Http\Middleware\TrackInactivity::handle

[View full stacktrace](https://app.bugsnag.com/vatsim-uk/core/errors/5f1624fede15e000188b758e?event_id=5f1624fe005b890cb9da0000&i=gh&m=ci)

*Created by Calum Towers via Bugsnag*

PR #1871, opened 2020-08-01
Properly Catch UKCP FileNotFound Exceptions

CalumTowers:
Fixed #1858 

The `Storage` facade uses League's Flysystem so we need to catch that exception.


CalumTowers:
Fixed #1858

PR #1889, opened 2020-08-10
1830 - Throttle Secondary Forum Groups

### LINKED ISSUE #1830: Improve IssueSecondaryForumGroup Performance
## Error in Core

**IPS\Db\Exception** in **App\Jobs\ExternalServices\IssueSecondaryForumGroup**
MySQL server has gone away

[View on Bugsnag](https://app.bugsnag.com/vatsim-uk/core/errors/5ee1ac589d70ef0018116e49?event_id=5f0a8ad0005b89e778fd0000&i=gh&m=ci)

## Stacktrace

    app/Jobs/ExternalServices/IssueSecondaryForumGroup.php:42 - App\Jobs\ExternalServices\IssueSecondaryForumGroup::handle

[View full stacktrace](https://app.bugsnag.com/vatsim-uk/core/errors/5ee1ac589d70ef0018116e49?event_id=5f0a8ad0005b89e778fd0000&i=gh&m=ci)

*Created by Calum Towers via Bugsnag*

CalumTowers:
Hoping this resolves #1830


app/Jobs/ExternalServices/IssueSecondaryForumGroup.php
```diff
@@ -32,49 +34,53 @@ public function __construct(int $cid, int $forumGroup)
      */
     public function handle()
     {
-        $ipboard = new Ipboard();
+        Redis::throttle('issue-secondary-forum-group')->allow(15)->every(60)->then(function () {
```
AxonC, line 37: What is the 15 represent?

CalumTowers: 15 instances of this task can be ran every 60 seconds. It throttles the number of times we can physically run this job. It is designed to stop you hammering an API, but in our case, it should stop us hammer our own forum database.

AxonC: Can you put that into a semi-descriptive variable to improve readability? Something like `$maximumTaskAllowance = 15`

CalumTowers: I'm not sure that helps?
Its a documented feature here: https://laravel.com/docs/7.x/queues#rate-limiting

And it currently reads "throttle, allow 15, every 60" which seems quite readable to me?

AxonC: 6 vs half-dozen. I'd like to be able to read that at first glance without needing the docs. I'm easy either way, but I didn't know what it was till I read the docs myself.

CalumTowers: I think extracting a var out for this feels a bit odd. I'd rather stick to convention that is outlined in the docs

A-Lawrence: Can I wade in? Why don't you extract this whole thing into a middleware?

`<?php

namespace App\Jobs\Middleware;

use Illuminate\Support\Facades\Redis;

class RateLimited
{
    /** @var null */
    private $key;

    /** @var int */
    private $allow;

    /** @var int */
    private $every;

    /** @var int */
    private $retryAfter;

    public function __construct(string $key = null, ?int $allow = 10, ?int $every = 60, ?int $retryAfter = 10)
    {
        $this->key        = $key ?? 'rate_limited_queue_job';
        $this->allow      = $allow ?? 10;
        $this->every      = $every ?? 60;
        $this->retryAfter = $retryAfter ?? 10;
    }

    public function handle($job, $next)
    {
        Redis::throttle($this->key)
             ->allow($this->allow)
             ->every($this->every)
             ->then(function () use ($job, $next) {
                 // Lock obtained...

                 $next($job);
             }, function () use ($job) {
                 // Could not obtain lock...

                 $job->release($this->retryAfter);
             });
    }
}`

Then you can name your constructor argument whatever you want, without needing to repeat your throttling on every job => https://laravel.com/docs/7.x/queues#job-middleware

CalumTowers: Good shout! Thanks 👍

PR #1893, opened 2020-08-11
fix: permission name cannot be null, error

A-Lawrence:
In the latest version of Laravel 7.25, there are issues with the Spatie permission package, because of changes made to how $guarded works (in light of a security issue).  This should resolve those issues until the package is updated.

cc: @CalumTowers <- he asked me to open this PR.

### DIFF
--- a/app/Models/Permission.php
+++ b/app/Models/Permission.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class Permission extends \Spatie\Permission\Models\Permission
+{
+    protected $guarded = [];
+    protected $fillable = ['name', 'guard_name'];
+}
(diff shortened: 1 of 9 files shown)


A-Lawrence:
> Thanks @A-Lawrence 👀
> 
> Any chance you could fix the references the Spatie Permission Model in the migrations so that they run correctly (and the tests pass on GitHub Actions)? 👼

...you don't want much, do you? I thought I was demonstrating the fix 😂

AxonC:
> > Thanks @A-Lawrence 👀
> 
> > 
> 
> > Any chance you could fix the references the Spatie Permission Model in the migrations so that they run correctly (and the tests pass on GitHub Actions)? 👼
> 
> 
> 
> ...you don't want much, do you? I thought I was demonstrating the fix 😂 

Cheek of it...

PR #1895, opened 2020-08-13
Discord Invite Email

CalumTowers:
Discord invite email, will be used to manually send to people that have joined us on Slack but not Discord, but may be useful in future too.

image

### DIFF
--- a/app/Notifications/Mship/DiscordInvite.php
+++ b/app/Notifications/Mship/DiscordInvite.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace App\Notifications\Mship;
+
+use App\Models\Mship\Account;
+use App\Notifications\Notification;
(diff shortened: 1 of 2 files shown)


app/Notifications/Mship/DiscordInvite.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Get the notification's delivery channels.
+     *
+     * @param  mixed $notifiable
+     * @return array
+     */
+    public function via($notifiable)
+    {
+        return ['mail', 'database'];
```
AxonC, line 24: if you're writing to the db, I think you need an array representation of the notification

CalumTowers: There is, its just an empty array (similar to some of the other notifications we have, it just doesn't save any attributes)

PR #1897, opened 2020-08-14
fix: CERT Update update frequency

### LINKED ISSUE #1896: CERT User Updates
We currently have a lot of users in our database, and our monthly processing of people who have not logged in for a while generally can't keep up with processing changes and things such as GDPR removals.

We should look to:
- Alter the Members:CertUpdate command to change which users are targeted
- Explore options to actively delete users with: no login for an extended period of time (years), have no relationship / dependant data, and are not division members (this might be best accomplished as a standalone script that is not in core)

https://github.com/VATSIM-UK/core/blob/master/app/Console/Commands/Members/UpdateMembers.php

atoff:
Does the following:
- Add rate middleware to the UpdateMember job to limit to 100 updates per 60 seconds
- Reinstates the hourly Cert Update Job
- Increases Monthly update frequency to tri-monthly
- Increase the capacity of the all-members job

Closes #1896

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -53,6 +53,10 @@ class Kernel extends ConsoleKernel
 
         // === By Hour === //
 
+        $schedule->command('members:certupdate', ['--type=hourly'])
+            ->hourly()
+            ->runInBackground();
+
         $schedule->command('members:certimport')
             ->cron('30 */2 * * *') // every second hour
             ->runInBackground();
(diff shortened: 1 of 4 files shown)


atoff:
@CalumTowers thoughts / opinions? This should clear our backlog over 3 and a bit months, or we could manually run some monthly cert updates to supplement this, but I reckon this should be a pretty good schedule for us

PR #1942, opened 2020-10-04
feat: Remove unused tables and models

### LINKED ISSUE #1902: Remove Unused Tables
There are a number of tables that are no longer used and we should drop. Associated models can also be removed.

- basic_users (was only used for Sendgrid & Mailgun, check if required elsewhere)
- community_group
- community_membership
- email_events
- jobs_pending (now using Redis)
- messages_thread
- messages_thread_participant
- messages_thread_post
- short_url (functionality was moved to vats.im)
- staff_account_position
- staff_attribute_position
- staff_attributes
- staff_positions
- staff_services
- stations (? check for uses)
- statistic (also remove from scheduler)
- sys_config (only had one use and is no longer needed)
- sys_timeline_action
- sys_timeline_entry

atoff:
Closes #1902


AxonC (dismissed):
Worried that our tests still pass with a mass cull but nothing stands out to me

atoff:
> Worried that our tests still pass with a mass cull but nothing stands out to me

A hopeful sign that they are indeed unused :P

CalumTowers:
This looks good to me. Before merging I think it would be sensible to deploy this to beta and get people to do some clicking around - make sure we haven’t missed anything. Given the majority of this was entirely untested I think I need some more assurance before releasing it to prod.

@AxonC Thoughts?

AxonC:
> 
> 
> This looks good to me. Before merging I think it would be sensible to deploy this to beta and get people to do some clicking around - make sure we haven’t missed anything. Given the majority of this was entirely untested I think I need some more assurance before releasing it to prod.
> 
> @AxonC Thoughts?

Yep, good idea to me.

PR #1975, opened 2020-11-14
1957 - Laravel 8 Upgrade

### LINKED ISSUE #1957: Laravel 8 Upgrade

CalumTowers:
Resolves #1957


atoff (approved):

app/Models/Concerns/TracksChanges.php
```diff
@@ -70,8 +70,6 @@ public function trackChanges()
         $attributes = $this->originalTracked;
 
         foreach ($this->trackableFromArray($attributes) as $key => $value) {
-            $key = $this->removeTableFromKey($key);
```
atoff, line 73: Is this related to the update? I don't see this method in the file at all, not sure where it comes from

CalumTowers: It has been deprecated for a while (ref: https://github.com/laravel/framework/pull/33865) and was removed in v8.0

CalumTowers: Just double-tripe checked.
The method has, for a while, just simply returned the `$key` unchanged. It used to do the below;

```php
        if (! Str::contains($key, '.')) {
            return $key;
        }

        return last(explode('.', $key));
```

AxonC (approved):
Pleasing

PR #1976, opened 2020-11-14
1944 - CERT Sync Schedule Change

### LINKED ISSUE #1944: Review CERT Update Schedule
Review the batches in which we update users from CERT via the scheduler to ensure we capture people as often as possible without a negative impact on performance.

CalumTowers:
The automated updates from CERT (using `Members:CertUpdate`) are currently flawed in a number of ways.

- It primarily focuses on users that have logged in, despite users being updated when they login automatically, thus not needing to be synced by this command
- It limits the number of members being captured that have never logged in by capping it at 1000 people every 30 days. This results in us holding outdated data on members as we never get a chance to clear the backlog
- It has very little chance to update inactive members - again, resulting in outdated data (e.g. for right to be forgotten requests)

My proposal, and contained within this PR, is that we process updates once per hour, the same way each time. Each run is capped at 1000 members and queued on a dedicated queue as it is now.
The 1000 members are selected only using their `cert_checked_at`. By sorting it `ASC`, we prioritise those that have never been checked (`null`) and then in reverse date order to get the oldest first.

Given the number of members we currently have in production, and assuming no one logs in to prompt an update on the fly, this would ensure that every user in our database "phones home" to .net at least once every 7 days.

Once merged, this resolves #1944

### DIFF
--- a/app/Console/Commands/Members/UpdateMembers.php
+++ b/app/Console/Commands/Members/UpdateMembers.php
@@ -18,8 +18,7 @@ class UpdateMembers extends Command
      * @var string
      */
     protected $signature = 'Members:CertUpdate
-                        {max_members=1000}
-                        {--t|type=all : Which update are we running? Hourly, Daily, Weekly or Monthly?}
+                        {max_members=5000}
                         {--f|force= : If specified, only this CID will be checked.}';
 
     /**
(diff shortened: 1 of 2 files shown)


A-Lawrence:
To add some context, historically there were ridiculous limitations placed on us as to the number of updates permitted from a single IP in a 24 hour period, from dot net (the suggestions of webhooks to avoid all these calls was laughed at). Provided these limitations have been removed, then it LGTM!

atoff (approved):
Makes sense 👍

PR #1995, opened 2020-12-04
Upgrade Data Feed

CalumTowers:


app/Console/Commands/NetworkData/ProcessNetworkData.php
```diff
@@ -178,20 +175,20 @@ private function processPilots()
...
-                'cruise_altitude'     => $pilotData['planned_altitude'],
-                'cruise_tas'          => $pilotData['planned_tascruise'],
-                'route'               => $pilotData['planned_route'],
-                'remarks'             => $pilotData['planned_remarks'],
+                'alternative_airport' => $pilotData['flight_plan']['alternate'],
+                'aircraft'            => $pilotData['flight_plan']['aircraft'],
+                'cruise_altitude'     => $pilotData['flight_plan']['altitude'],
+                'cruise_tas'          => $pilotData['flight_plan']['cruise_tas'],
+                'route'               => $pilotData['flight_plan']['route'],
+                'remarks'             => $pilotData['flight_plan']['remarks'],
```
AxonC, line 191: Maybe worth defining a 'getFlightplanField' method instead of getting the flight plan key each time?

CalumTowers: I think the extraction could be overkill - my goal is to remove the need for this entirely and start using the API to grab the stats we need rather than storing it ourselves

PR #1999, opened 2020-12-11
Console / Network Data Improvements

CalumTowers:
- Disables Discord Manager whilst we try and improve performance
- Extends grace period for network data
- Adds error handling for when the network data feed is unavailable
- Improves output of the `networkdata:download` command to show progress
- Small refactors within the network data command

### DIFF
--- a/app/Console/Commands/NetworkData/ProcessNetworkData.php
+++ b/app/Console/Commands/NetworkData/ProcessNetworkData.php
@@ -149,10 +149,12 @@ class ProcessNetworkData extends Command
             $progressBar->advance();
         }
 
-        $this->endExpiredAtcSessions($awaitingUpdate);
         $progressBar->finish();
         $this->newLine();
 
+        if($awaitingUpdate->isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


app/Console/Commands/NetworkData/ProcessNetworkData.php
```diff
@@ -54,11 +54,20 @@ public function __construct()
      */
     public function handle()
     {
+        $this->info('Getting network data from VATSIM.');
+
+        if ($this->networkData->failed() || ! $this->networkData->json()) {
+            $this->error('VATSIM feed unavailable.');
+            exit();
```
AxonC, line 61: Would it be simpler to just return?

AxonC (approved):

app/Console/Commands/NetworkData/ProcessNetworkData.php
CalumTowers: This was me trying to be super explicit as to the fact that we are killing the process there and then.

PR #2006, opened 2020-12-30
Fix Non-P0 users being synced to external services

atoff:
Fixes the fully defined function to only report true if the user has a valid ATC qual (OBS+)


CalumTowers (approved):
👍

CalumTowers (approved):

PR #2013, opened 2021-01-09
Fix Unauthenticated Redirect

atoff:


CalumTowers (commented):
Thanks @atoff. Looking at it now, I'm wondering whether the route name needs to be login rather than dashboard so it's explicit?

atoff:
> Thanks @atoff. Looking at it now, I'm wondering whether the route name needs to be login rather than dashboard so it's explicit?

We can't name it login without changing the route name of the "actual" login route. Could rename to "landing" as that might make more sense>

CalumTowers:
> > Thanks @atoff. Looking at it now, I'm wondering whether the route name needs to be login rather than dashboard so it's explicit?
> 
> 
> 
> We can't name it login without changing the route name of the "actual" login route. Could rename to "landing" as that might make more sense>

Agree that makes sense

CalumTowers (approved):

PR #2021, opened 2021-01-17
feat: Sync member bans to discord

AxonC:
Allows a 'suspended member' role to be added to a Discord user when discovered by
the discord manager command.

This role is configured via environment variables.﻿

I modified the accessibility to the grantRoels() and removeRoles() methods to a) assist in testing and I didn't see a need to have an access modifier because it only depends on things injected to the constructor.


tests/Unit/Command/DiscordManagerTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->mockRoleId = 122904438483;
+
+        Config::set('services.discord.suspended_member_role_id', $this->mockRoleId);
+
+        // discord is is not relevant; just that it is registered for completeness.
```
CalumTowers, line 29:
```suggestion
        // Discord id is not relevant; just that it is registered for completeness.
```

app/Console/Commands/ExternalServices/ManageDiscord.php
```diff
@@ -100,13 +108,17 @@ protected function grantRoles()
...
-    protected function removeRoles()
+    public function removeRoles()
     {
         $account = $this->account;
         $discord = $this->discord;
 
         $currentRoles = $discord->getUserRoles($account);
 
+        if (! $account->isBanned && $currentRoles->contains($this->suspendedRoleId)) {
+            $discord->removeRoleById($account, $this->suspendedRoleId);
```
CalumTowers, line 119: Could we somehow, within this block, also remove any other roles that they have in `$currentRoles`? Regardless of what `DiscordRole` says they should have?

We wouldn't need to worry about giving them the roles back after being unbanned as that would be picked up by lines 101 onwards?

CalumTowers (changes requested):
We'll also need to add this command back to the scheduler, running once per day around midday I'd suggest.

app/Console/Commands/ExternalServices/ManageDiscord.php
CalumTowers: Actually, this would be in the `grantRoles` block, perhaps by calling a `$this->removeAllRoles()` method?

app/Console/Commands/ExternalServices/ManageDiscord.php
```diff
@@ -83,13 +87,17 @@ protected function assignNickname()
...
 
-    protected function grantRoles()
+    public function grantRoles()
     {
         $account = $this->account;
         $discord = $this->discord;
 
         $currentRoles = $discord->getUserRoles($account);
 
+        if ($account->isBanned && ! $currentRoles->contains($this->suspendedRoleId)) {
```
CalumTowers, line 97: Could we somehow, within this block, also remove any other roles that they have in $currentRoles? Regardless of what DiscordRole says they should have?

Perhaps just by extracting that logic (of giving the suspended role and removing all others) to a `$this->processSuspendedMember()` method and not letting them get to the `grantRoles()` or `removeRoles()` methods?

CalumTowers (approved):

CalumTowers (approved):

PR #2026, opened 2021-01-21
Fix - Primary Qualification Selection

atoff:
Selects the primary qualification based on the "vatsim" code, rather than date added


CalumTowers (approved):
👍

PR #2028, opened 2021-01-21
Feat - Utilise .NET Auth required_scopes

atoff:
Utilise required_scopes on OAuth requests to reduce chance of user unselecting and then getting an error message.
Disables the ability to deselect the requested scopes.
image


CalumTowers (approved):
👍

PR #2032, opened 2021-01-25
Update staff tables

bbrushfield:
include new/old changes in position
Change Darren Hill to Training Manager and remove Ben Cook


CalumTowers (approved):
Thanks @bbrushfield 👍

PR #2036, opened 2021-01-28
Remove _R from approach

grahamdrabble:
EGNO_R_APP isn't mentioned in the change note. I've assumed that the military use of _R_APP will follow the same convention.....

### DIFF
--- a/resources/views/site/operations/sectors.blade.php
+++ b/resources/views/site/operations/sectors.blade.php
@@ -68,7 +68,7 @@
 
                     <ul>
                         <li>
-                            <strong>EGAA_R_APP</strong> - Belfast/Aldergrove (EGAA) and Belfast/City (EGAC)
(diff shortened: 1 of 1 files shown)


CalumTowers:
@jackedwards815 - Could you take a look at this?

jackedwards815 (commented):
The changes are correct, but only cover a small portion of the instances of R_APP

CalumTowers:
@grahamdrabble Would you mind updating all affected positions here please?

hsugden:
> @grahamdrabble Would you mind updating all affected positions here please?

These are the only instances of _R_ callsigns on this page, so the suggested changes are fine :)

jackedwards815:
We could do it just for this page, but ignores the wider issue of positions still with R_APP in them on the airports pages (currently hidden because of this).

PR #2038, opened 2021-01-28
Update Staff Page

LouisHacking:
Informed by Matthew Wilson - He left the role of 'initial flight instructor' ~6 months ago.


LouisHacking:
Accidentally added multiple commits, sorry @CalumTowers. You already know I'm a newb!

CalumTowers (changes requested):
Thanks @LouisHacking 
Until we consider altering the numbering (e.g. for VATUK 1 - 9) then I'd prefer to keep the order as it is.
Can we adjust this PR to just amend Matthew's role?

CalumTowers (approved):

PR #2039, opened 2021-02-01
feat: Sync Division Members Daily

atoff:
Not too sure what we should limit this job to by default - what do you think @CalumTowers?


CalumTowers:
@atoff We'll sync all of them. We'll see how the queue copes with it I guess...

CalumTowers (approved):

PR #2090, opened 2021-03-09
Home Page, Area sectors button

kye-taylor:
Adds a button on the hope page, that directly links viewers to the area sectors page.


CalumTowers:
This could be added to the navigation but I don't think it is suitable to add an additional button in this position on the page.

PR #2106, opened 2021-03-19
style: Change "Invisibility" to "Show as Online"

cocoiadrop:
This better reflects what the feature actually does (based on what I was told in #web-help), seeing as invisibility could be interpreted for something else.
Added tooltip to text to explain it, and swapped the Enable and Disable states so it actually makes sense.


CalumTowers:
This feature is isolated to our forum and should likely be made clear - they wouldn't be made invisible across all of our service. NPTM at this time.

PR #2122, opened 2021-04-03
Auto Sync Role Changes

CalumTowers:
This PR;

- Introduces two new events that fire when a role is assigned or removed from a user
- Adds the two new events to the SyncSubscriber so that when a role is added or removed, it syncs the user to external services
- Adds some missing tests for Discord to assert that the job was added to the queue

### DIFF
--- a/app/Events/Mship/Roles/RoleAssigned.php
+++ b/app/Events/Mship/Roles/RoleAssigned.php
@@ -11,11 +11,9 @@ class RoleAssigned extends Event
 {
     use SerializesModels;
 
-    /* @var Account */
-    public $account;
+    public Account $account;
 
-    /* @var Role */
-    public $role;
+    public Role $role;
 
     public function __construct(Account $account, Role $role)
     {
--- a/app/Events/Mship/Roles/RoleAssigned.php
+++ b/app/Events/Mship/Roles/RoleAssigned.php
@@ -4,8 +4,8 @@ namespace App\Events\Mship\Roles;
 
 use App\Events\Event;
 use App\Models\Mship\Account;
-use Spatie\Permission\Models\Role;
 use Illuminate\Queue\SerializesModels;
+use Spatie\Permission\Models\Role;
 
 class RoleAssigned extends Event
 {
--- a/app/Events/Mship/Roles/RoleAssigned.php
+++ b/app/Events/Mship/Roles/RoleAssigned.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace App\Events\Mship\Roles;
+
+use App\Events\Event;
+use App\Models\Mship\Account;
+use Spatie\Permission\Models\Role;
+use Illuminate\Queue\SerializesModels;
+
+class RoleAssigned extends Event
+{
+    use SerializesModels;
+
+    /* @var Account */
+    public $account;
+
+    /* @var Role */
+    public $role;
+
+    public function __construct(Account $account, Role $role)
+    {
+        $this->account = $account;
+        $this->role = $role;
+    }
+}
--- a/app/Events/Mship/Roles/RoleRemoved.php
+++ b/app/Events/Mship/Roles/RoleRemoved.php
@@ -4,8 +4,8 @@ namespace App\Events\Mship\Roles;
 
 use App\Events\Event;
 use App\Models\Mship\Account;
-use Spatie\Permission\Models\Role;
 use Illuminate\Queue\SerializesModels;
+use Spatie\Permission\Models\Role;
 
 class RoleRemoved extends Event
 {
--- a/app/Events/Mship/Roles/RoleRemoved.php
+++ b/app/Events/Mship/Roles/RoleRemoved.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace App\Events\Mship\Roles;
+
+use App\Events\Event;
+use App\Models\Mship\Account;
+use Spatie\Permission\Models\Role;
+use Illuminate\Queue\SerializesModels;
+
+class RoleRemoved extends Event
+{
+    use SerializesModels;
+
+    /* @var Account */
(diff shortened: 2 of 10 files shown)


app/Events/Mship/Roles/RoleAssigned.php
```diff
@@ -0,0 +1,25 @@
...
+use App\Models\Mship\Account;
+use Illuminate\Queue\SerializesModels;
+use Spatie\Permission\Models\Role;
+
+class RoleAssigned extends Event
+{
+    use SerializesModels;
+
+    /* @var Account */
+    public $account;
```
AxonC, line 15:
```suggestion
    public Account $account;
```
(Instead of the doc blocks, apply to file)

app/Events/Mship/Roles/RoleRemoved.php
```diff
@@ -0,0 +1,25 @@
...
+use App\Models\Mship\Account;
+use Illuminate\Queue\SerializesModels;
+use Spatie\Permission\Models\Role;
+
+class RoleRemoved extends Event
+{
+    use SerializesModels;
+
+    /* @var Account */
+    public $account;
```
AxonC, line 15: Same here on the typed properties

app/Models/Mship/Concerns/HasDiscordAccount.php
```diff
@@ -17,6 +17,10 @@ trait HasDiscordAccount
      */
     public function syncToDiscord()
     {
+        if (! config('services.discord.token')) {
+            return;
```
AxonC, line 21: Possibly worth logging something to Bugsnag or logging a critical error?

tests/Unit/Account/Sync/RoleSyncTest.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        Queue::fake();
+
+        event(new RoleAssigned($this->user, $this->role));
+
+        Queue::assertPushed(SyncToCTS::class, 1);
+        Queue::assertPushed(SyncToMoodle::class, 1);
+        Queue::assertPushed(SyncToHelpdesk::class, 1);
+        Queue::assertPushed(SyncToDiscord::class, 1);
+        //Queue::assertPushed(SyncToForums::class, 1);
```
AxonC, line 61: Rogue commented out code

tests/Unit/Account/Sync/RoleSyncTest.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        Queue::fake();
+
+        event(new RoleRemoved($this->user, $this->role));
+
+        Queue::assertPushed(SyncToCTS::class, 1);
+        Queue::assertPushed(SyncToMoodle::class, 1);
+        Queue::assertPushed(SyncToHelpdesk::class, 1);
+        Queue::assertPushed(SyncToDiscord::class, 1);
+        //Queue::assertPushed(SyncToForums::class, 1);
```
AxonC, line 75: Ditto here

app/Models/Mship/Concerns/HasRoles.php
```diff
@@ -0,0 +1,75 @@
...
+        $this->fireRoleRemovedEvent($role);
+
+        return $this;
+    }
+
+    /**
+     * @param $role
+     * @return bool
+     */
+    public function fireRoleRemovedEvent($role)
```
AxonC, line 65: There's a school of thought here that might say that passing an array of roles is simpler, even if it has only one element, and then just looping through the roles passed.

AxonC (changes requested):

tests/Unit/Account/Sync/RoleSyncTest.php
CalumTowers: Intentionally disabled, has been for some time. Forum doesn't sync on demand, it syncs every member overnight via the scheduler. @atoff is on the case to enable it on demand via the SyncSubscriber.

tests/Unit/Account/Sync/RoleSyncTest.php
CalumTowers: Intentionally disabled, has been for some time. Forum doesn't sync on demand, it syncs every member overnight via the scheduler. @atoff is on the case to enable it on demand via the SyncSubscriber.

app/Models/Mship/Concerns/HasDiscordAccount.php
CalumTowers: Reality is this is only ever going to happen in testing. Its a cheap and lazy way of avoiding the actual code executing in the test suite rather than mocking it. FWIW, the same approach is done for CTS, Moodle and Helpdesk (only they check for the presence of a matching oauth client).

app/Events/Mship/Roles/RoleAssigned.php
CalumTowers: c7744ff3

app/Events/Mship/Roles/RoleRemoved.php
CalumTowers: c7744ff3

app/Models/Mship/Concerns/HasRoles.php
CalumTowers: We can't really. The package supports either (ref: https://spatie.be/docs/laravel-permission/v4/basic-usage/role-permissions);

```
$account->assignRole('role_one');
$account->assignRole('role_one', 'role_two');
$account->assignRole(['role_one', 'role_two']);
```

If I modified this method to only support arrays, the first two approaches wouldn't work.

I actually grabbed this code from one of the Spatie maintainers who had commented on a GH issue on the package asking for this to be added. Doesn't look like they planned to add it so they suggested this.

AxonC (approved):

PR #2136, opened 2021-04-15
fix: Cert Request Check Cooldown

atoff:
Having recently used the cache for something else, I release now that the Cache TTL operates in seconds rather than ms.
Therefore this is currently 1000 hours!
Changed to 60 seconds * 60 i.e. an actual hour!

### DIFF
--- a/app/Http/Controllers/Mship/Management.php
+++ b/app/Http/Controllers/Mship/Management.php
@@ -89,7 +89,7 @@ class Management extends \App\Http\Controllers\BaseController
                 ->withError('Emails entered are different.  You need to enter the same email, twice.');
         }
 
-        if (!$this->account->hasEmail($email)) {
+        if (! $this->account->hasEmail($email)) {
             $this->account->addSecondaryEmail($email);
         } else {
             return Redirect::route('mship.manage.dashboard')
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Mship/Management.php
```diff
@@ -264,7 +264,7 @@ public function getVerifyEmail($code)
 
     public function requestCertCheck()
     {
-        $ranRecently = ! Cache::add('USER_REQUEST_CERTCHECK_'.Auth::user()->id, '1', 1 * 60 * 60 * 1000); // 1 hour
+        $ranRecently = ! Cache::add('USER_REQUEST_CERTCHECK_'.Auth::user()->id, '1', 1 * 60 * 60); // 1 hour
```
CalumTowers, line 267:
```suggestion
        $ranRecently = ! Cache::add('USER_REQUEST_CERTCHECK_'.Auth::user()->id, '1', now()->addHour()); // 1 hour
```
Something like (e.g. passing in a carbon instance) may be clearer to avoid similar confusion in future?

CalumTowers (changes requested):

PR #2139, opened 2021-04-19
feat: On-demand Forum Sync

atoff:
Few other areas we should probably fix / decide if we want to convert whilst here:
- SyncMentors
- Remove SyncCommunity command?
- SyncBanToForum (Might be able to remove now?)
- IssueSecondaryForumGroup


atoff:
@CalumTowers I'm thinking that changing the other things that use the IPB classes is going to be quite a large refactor job. This PR is ready to go so I suggest we implement this and look to phase out our usage of the IPB classes later on (the main problem is that some of these are using that `vatsim_id` column on the members table which isn't ideal..)

app/Models/Mship/Concerns/HasForumAccount.php
```diff
@@ -4,52 +4,50 @@
...
+        if (! $ipsAccount) {
+            // No user. Abort;
+            return;
+        }
+
+        // Set data
+        DB::table("{$communityDb}.ibf_core_members")
+            ->where('member_id', $ipsAccount->member_id)
+            ->update([
+                'name' => $this->real_name,
```
CalumTowers, line 39: If we use `$this->name` we will honour any nicknames that are set (which I think is worth doing)

CalumTowers (changes requested):
Happy with that - just one subtle change before merging!

CalumTowers (approved):

PR #2151, opened 2021-04-25
Handle Discord Name Limit

CalumTowers:
Discord limits nicknames to a maximum of 32 characters. In the event that a full name (e.g. `name_first` and `name_last`) is greater than 32 characters, the API will throw an exception when attempting to set the nickname.

This PR;

- Introduces a method to get the user's first name; either the set `nickname` or `name_first`. This is accessed via `$account->firstName`.
- Refactors the `$account->name` to defer to the new `$account->firstName` to get either the `nickname` or `name_first`.
- Adds a method to the `HasDiscordAccount` trait to get the user's "Discord name". This will be their full name (either nickname or first name, space, then surname) or, in the event that this would be over 32 characters, it will replace their surname with their initial.

The hole in this PR will be that if the user has a long first name then changing their surname could still render them over 32 characters. Having reviewed our member list, I don't believe this will be an issue for now.

In the future, we should afford members the flexibility to change their nickname. More thought required here.

### DIFF
--- a/app/Models/Mship/Account.php
+++ b/app/Models/Mship/Account.php
@@ -383,17 +383,23 @@ class Account extends Model implements AuthenticatableContract, AuthorizableCont
     /**
      * Get the user's full name.
      *
-     * If a nickname is set, that will be used in place of name_first.
-     *
      * @return mixed|string
      */
     public function getNameAttribute()
     {
-        if ($this->nickname != null) {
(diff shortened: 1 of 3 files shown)


atoff (dismissed):
All looks fine, only thing I would consider is not calling the new attribute `first_name` - this could lead to some confusion with `name_first`. Prehaps `name_first_preferred` or `preferred_first_name`?

PR #2161, opened 2021-05-05
fix(schedule): Allow overlapping on network data download tasks

AndyTWF:


app/Console/Kernel.php
```diff
@@ -37,8 +37,7 @@ protected function schedule(Schedule $schedule)
 
         $schedule->command('networkdata:download')
             ->everyTwoMinutes()
-            ->graceTimeInMinutes(10)
-            ->withoutOverlapping();
```
CalumTowers, line 41: Can we do the same as UKCP in here and pass in 5?

CalumTowers (changes requested):

PR #2230, opened 2021-07-09
Migrate from VATSIM XML to New API

CalumTowers:
Need to sort pagination on the ImportMembers script

### DIFF
--- a/app/Console/Commands/Members/ImportMembers.php
+++ b/app/Console/Commands/Members/ImportMembers.php
@@ -78,7 +78,7 @@ class ImportMembers extends Command
         };
 
         // TODO: possibly add some OhDear functionality if this request fails?
-        $url = config('vatsim-api.base') . 'divisions/GBR/members';
+        $url = config('vatsim-api.base').'divisions/GBR/members';
         $response = Http::withToken(config('vatsim-api.key'))->get($url)->json();
 
         $memberCollection = collect();
--- a/app/Console/Commands/Members/ImportMembers.php
+++ b/app/Console/Commands/Members/ImportMembers.php
@@ -78,7 +78,7 @@ class ImportMembers extends Command
         };
 
         // TODO: possibly add some OhDear functionality if this request fails?
-        $url = config('vatsim-api.base') . 'divisions/GBR/members';
+        $url = config('vatsim-api.base').'divisions/GBR/members';
         $apiToken = config('vatsim-api.key');
         $response = Http::withHeaders([
             'Authorization' => "Token {$apiToken}",
--- a/app/Console/Commands/Members/ImportMembers.php
+++ b/app/Console/Commands/Members/ImportMembers.php
@@ -2,15 +2,17 @@
 
 namespace App\Console\Commands\Members;
 
-use App\Console\Commands\Command;
-use App\Libraries\AutoTools;
-use App\Models\Mship\Account;
-use App\Models\Mship\Qualification;
-use App\Models\Mship\State;
-use App\Notifications\Mship\WelcomeMember;
 use DB;
 use Exception;
+use App\Models\Mship\State;
+use App\Libraries\AutoTools;
+use App\Models\Mship\Account;
+use Illuminate\Support\Carbon;
+use App\Console\Commands\Command;
+use App\Models\Mship\Qualification;
+use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Validator;
+use App\Notifications\Mship\WelcomeMember;
 
 /**
  * Utilizes the CERT divdb file to import new users and update existing user emails.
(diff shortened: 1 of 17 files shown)


composer.json
```diff
@@ -49,10 +49,6 @@
         }
     ],
     "repositories": [
-        {
-            "type": "vcs",
```
AxonC, line 53: You'll need to remove line 103 from this file too

app/Console/Commands/Members/ImportMembers.php
```diff
@@ -60,6 +56,41 @@ public function handle()
...
+                'cid' => $result['id'],
+                'rating_atc' => $result['rating'],
+                'rating_pilot' => $result['pilotrating'],
+                'name_first' => $result['name_first'],
+                'name_last' => $result['name_last'],
+                'email' => $result['email'],
+                'age_band' => $result['age'],
+                'city' => $result['countystate'],
+                'country' => $result['country'],
+                'experience' => '',
```
AxonC, line 83: Is there an argument to say these columns with empty strings can be dropped from our own DB whilst we are making these changes? 

As a separate point, do we need all of these fields?

app/Console/Commands/Members/ImportMembers.php
```diff
@@ -113,24 +144,28 @@ protected function createNewMember($member_data)
...
-            }
-
-            if (isset($_prevRat->PreviousRatingInt)) {
-                $prevAtcRating = Qualification::parseVatsimATCQualification($_prevRat->PreviousRatingInt);
-
-                if ($prevAtcRating) {
-                    $member->addQualification($prevAtcRating);
-                }
-            }
+            // This user has an admin rating but there is currently no support
```
AxonC, line 147: This worries me a little

app/Console/Commands/Members/ImportMembers.php
```diff
@@ -3,15 +3,15 @@
 namespace App\Console\Commands\Members;
 
 use App\Console\Commands\Command;
-use App\Libraries\AutoTools;
```
AxonC, line 6: Can this file be deleted?

app/Console/Commands/Members/ImportMembers.php
CalumTowers: My aim with this PR is to match exactly what the AutoTools stuff was giving us so we aren't rewriting the entire script. 

If we get the chance to rewrite it then I totally agree we need to review things like what data we have. 

For what it's worth I don't actually think we do anything with this but I wanted to keep the structure exactly as is to be safe.

app/Console/Commands/Members/ImportMembers.php
CalumTowers: I'm not convinced prat has been working anyway to be fair. I've raised it to .net to consider for a future update.

app/Console/Commands/Members/ImportMembers.php
CalumTowers: It will be able to but I want to keep it in just for now in case we come to migrate anything like prat and want to reference it in the future.

CalumTowers:
@AxonC Will you get a chance to look at the pagination stuff or want to leave it for me when I’m back up and running?

AxonC:
> @AxonC Will you get a chance to look at the pagination stuff or want to leave it for me when I’m back up and running?

I will need a token, but there isn't anything about pagination in the API docs

app/Providers/AppServiceProvider.php
```diff
@@ -103,5 +104,24 @@ public function registerValidatorExtensions()
...
+            Helper macro to configure an API request to the VATSIM API
+
+            $path: path after the last slash in the configured API URL
+            $includeBase: whether or not the base path should be used when constructing the URL
+            $method: type of HTTP request to make. Defaults to GET.
+
+            TODO: Open up possibility of sending post request with body.
+            TODO: When using PHP8, named_arguments would make this much nicer to use from a developer perspective.
+        */
+        Http::macro('vatsimAPIRequest', function ($path, $includeBase = true, $method = 'get') {
```
CalumTowers, line 118: @AxonC Can we refactor this macro out and keep it in line for now? Then we can deploy this I think

AxonC: Removed

CalumTowers:
@AxonC Does withToken work? I'm fairly sure that is for Bearer tokens but the VATSIM API uses Token rather than Bearer.

AxonC:
> @AxonC Does withToken work? I'm fairly sure that is for Bearer tokens but the VATSIM API uses Token rather than Bearer.

My bad, was working on the nexus stuff and this at the same time

CalumTowers:
> > @AxonC Does withToken work? I'm fairly sure that is for Bearer tokens but the VATSIM API uses Token rather than Bearer.
> 
> My bad, was working on the nexus stuff and this at the same time

This looks good to merge now, if you're happy hit approve (as it is my PR I can't).
Then I'll sort the deployment as we'll need to change some env vars.

AxonC (approved):

PR #2231, opened 2021-07-12
feat: Sync visiting controllrs to CTS when accepted

AxonC:
Not the finest bit of code, but it does appear to work. I don't have a full live CTS to test against locally, but this matches what data is expected in the production system for a visiting controller.

The logic for the internal IDs is straight from CTS. I dare not simplify it in case it is somehow messed up (there is a refactoring opportunity I can spot straight away with the boolean logic.). 

Also, there are often things not worth unit testing. I firmly put this in that category.

### DIFF
--- a/app/Models/Mship/Concerns/HasCTSAccount.php
+++ b/app/Models/Mship/Concerns/HasCTSAccount.php
@@ -47,7 +47,7 @@ use Illuminate\Support\Facades\DB;
             // if no CTS account exists, lets sync to create a new one, and not update anything this time round.
             if (! $ctsAccount) {
                 // for a division member, use the join timestamp of them joining, else use an empty timestamp.
-                // this is how CTS in its wisdom puts this column on visiting controllers.
+                // this is how CTS puts this column on visiting controllers.
                 $is_visitor = $this->primary_permanent_state->code != 'DIVISION';
                 $joined_div = ! $is_visitor
                     ? $this->primary_permanent_state->pivot->start_at
--- a/app/Models/Mship/Concerns/HasCTSAccount.php
+++ b/app/Models/Mship/Concerns/HasCTSAccount.php
@@ -44,6 +44,9 @@ use Illuminate\Support\Facades\DB;
 
             $ctsAccount = DB::table("{$ctsDatabase}.members")->where('cid', $this->id)->first();
 
+            // if user has nickname, prefer that name; else use full qualified name.
+            $name = $this->nickname ?: $this->real_name;
+
             // if no CTS account exists, lets sync to create a new one, and not update anything this time round.
             if (! $ctsAccount) {
                 // for a division member, use the join timestamp of them joining, else use an empty timestamp.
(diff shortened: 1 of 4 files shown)


app/Models/Mship/Concerns/HasCTSAccount.php
```diff
@@ -26,8 +44,33 @@ public function syncToCTS()
 
             $ctsAccount = DB::table("{$ctsDatabase}.members")->where('cid', $this->id)->first();
 
+            // if no CTS account exists, lets sync to create a new one, and not update anything this time round.
             if (! $ctsAccount) {
-                // No user exists. Abort.
+                // for a division member, use the join timestamp of them joining, else use an empty timestamp.
+                // this is how CTS in its wisdom puts this column on visiting controllers.
```
CalumTowers, line 50: Reword please!

app/Models/Mship/Concerns/HasCTSAccount.php
```diff
@@ -26,8 +44,33 @@ public function syncToCTS()
...
+                $is_visitor = $this->primary_permanent_state->code != 'DIVISION';
+                $joined_div = ! $is_visitor
+                    ? $this->primary_permanent_state->pivot->start_at
+                    : gmdate('Y-m-d H:i:s');
+
+                $newMember = [
+                    'old_rts_id' => 0,
+                    'id' => $this->generateCTSInternalID($this->id),
+                    'cid' => $this->id,
+                    'name' => $this->real_name,
```
CalumTowers, line 60: Does this honour nicknames?

app/Models/Mship/Concerns/HasCTSAccount.php
```diff
@@ -26,8 +44,33 @@ public function syncToCTS()
 
             $ctsAccount = DB::table("{$ctsDatabase}.members")->where('cid', $this->id)->first();
 
+            // if no CTS account exists, lets sync to create a new one, and not update anything this time round.
             if (! $ctsAccount) {
-                // No user exists. Abort.
+                // for a division member, use the join timestamp of them joining, else use an empty timestamp.
+                // this is how CTS in its wisdom puts this column on visiting controllers.
+                $is_visitor = $this->primary_permanent_state->code != 'DIVISION';
+                $joined_div = ! $is_visitor
```
CalumTowers, line 52: Double check the live CTS - I feel like this should be null if the member is visiting (as they've never actually joined us). If it's not nullable, I would use the state start_at for the visiting row as the joined at.

CalumTowers (changes requested):
Could you update the variables_with_underscores to be camelCase for consistency? Otherwise seems good to me!

app/Models/Mship/Concerns/HasCTSAccount.php
AxonC: From what I could find, it is literally this empty type timestamp (gets written as something like 00-00-0000 00:00:00). It's a non-nullable column with no default.

app/Models/Mship/Concerns/HasCTSAccount.php
AxonC: Probably not, but existing syncing via cron does not from what I've seen (happy to be wrong there)

atoff (dismissed):

app/Models/Mship/Concerns/HasCTSAccount.php
AxonC: Done :)

app/Models/Mship/Concerns/HasCTSAccount.php
CalumTowers: Let's honour nicknames as we do for other services like forum!

AxonC: Done! https://github.com/VATSIM-UK/core/pull/2231/commits/6499524184d03b87e6ba300ad45f727fc9c87018

CalumTowers (approved):

PR #2254, opened 2021-08-11
Import Member Re-Write

CalumTowers:
The current `ImportMember` script was disabled in production. Upon enabling it, I found a few issues and rather than digging into each one decided to try and clean it up a tad.

Given that we now have the `AccountAltered` event, the only thing this command needs to do is create the initial account with the minimum information required. Everything else (qualifications, states etc) will be handled by the subsequent run of the `SyncSubscriber`.

To use this command, you will need;

- A VATSIM API key for GBR
- To ensure that your database is running with the timezone of +00:00 (SELECT @@global.time_zone; / SET GLOBAL time_zone = '+00:00';) as some users have a registration date that is in a time that doesn't exist in BST (e.g. 01:05AM on the day clocks change)

The command will;

- Fetch all users from the VATSIM API that belong to GBR
- Honour the pagniation of this endpoint to ensure all users are obtained
- Handle any API downtime during the command running by exiting the while loop and proceeding with the members it has at the point the API goes down
- Validate the data for each member prior to proceeding to ensure the data is in the format we expect, skipping any that return unexpected data
- Create or update (depending on whether the id currently exists) an `Account`
- For newly created accounts, dispatch the `WelcomeMember` notification
- Not touch any qualifications, ratings or other data that is handled by the update process that follows a change in the account via model events
- Output helpful information on the progress of the command

The command is fairly slow on the first run on a fresh database as you are importing every single member. Subsequent runs that are mostly performing updates are much quicker.

### DIFF
--- a/app/Console/Commands/Members/ImportMembers.php
+++ b/app/Console/Commands/Members/ImportMembers.php
@@ -2,14 +2,14 @@
 
 namespace App\Console\Commands\Members;
 
-use App\Models\Mship\Account;
-use Illuminate\Support\Carbon;
 use App\Console\Commands\Command;
+use App\Models\Mship\Account;
+use App\Notifications\Mship\WelcomeMember;
+use Illuminate\Support\Carbon;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Validator;
-use App\Notifications\Mship\WelcomeMember;
-use Illuminate\Support\Collection;
 
 class ImportMembers extends Command
 {
@@ -65,7 +65,7 @@ class ImportMembers extends Command
 
         $response = Http::withHeaders([
             'Authorization' => "Token {$this->apiToken}",
-        ])->get(config('vatsim-api.base') . 'divisions/GBR/members');
+        ])->get(config('vatsim-api.base').'divisions/GBR/members');
 
         // Process first page of results
         foreach ($response->collect()->get('results') as $result) {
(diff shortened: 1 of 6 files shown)


app/Console/Commands/Members/ImportMembers.php
```diff
@@ -4,214 +4,102 @@
...
+            [
+                'name_first' => $member['name_first'],
+                'name_last' => $member['name_last'],
+                'email' => $member['email'],
+                'joined_at' => Carbon::create($member['reg_date']),
+                'inactive' => (int) $member['rating'] < 0,
+            ]
+        );
+
+        $account->wasRecentlyCreated ?? $account->notify(new WelcomeMember());
```
AxonC, line 102: Worth queueing this rather than dispatching the notification directly from the command?

app/Models/Mship/Concerns/HasQualifications.php
```diff
@@ -61,7 +61,7 @@ public function addQualification(Qualification $qualification)
      * @param int $atcRating The VATSIM ATC rating
```
AxonC, line 61: Docstrings don't reflect the optional param

AxonC (approved):

app/Console/Commands/Members/ImportMembers.php
CalumTowers: Its queued on the Notification class

app/Models/Mship/Concerns/HasQualifications.php
CalumTowers: 43e8d4fe

PR #2266, opened 2021-08-30
feat: Add base training place configuration and offer workflow

AxonC:
This PR addresses the basic needs of the training place automation workflow. PR was created at this stage to help with PR review
size going forward, and ensure bad decisions aren't taken early on in development.

Split into distinct sections to make clear what is implemented.

*Nova*:
- Adds TrainingPlace, TrainingPosition & Station resources, with relevant relationship form fields for WaitingList, Station and CTS Position
- Adds button to eligible Waiting List bucket to offer a training place to a user
- Adds modal component to provide dropdown for selecting available positions
- Removes style override in place for Feedback Message content - looking into replacing this with a package (https://github.com/tylernathanreed/nova-textwrap-field)

*Data Model & Process*:
- Adds tables and models for TrainingPlace, TrainingPosition, TrainingPlaceOffer
- Adds service for handling dispatching an offer to an account.
- Adds event framework for dispatching offer, designed to be consumed by several listeners for notifications etc.
- Adds method to TrainingPosition model to get available places for a given waiting list

*Implementation Notes*
- TrainingPlaceOffer uses a UUID as it will be used in future as a public-facing identifier linked
via email.
- TrainingPositions are linked to a cts position to facilitate automating granting permissions for a session.
This might also need to be linked to a CTS RTS ID in order to manage memberships

Note: some boilerplate might still be present in the code (e.g. notifications), thus why it is only going to a feature branch.

### DIFF
--- a/app/Models/Training/TrainingPlace/TrainingPosition.php
+++ b/app/Models/Training/TrainingPlace/TrainingPosition.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace App\Models\Training\TrainingPlace;
+
+use App\Models\Station;
+use App\Models\Training\WaitingList;
+use App\Models\Training\TrainingPlace;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\HasMany;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+
+class TrainingPosition extends Model
+{
+    use HasFactory;
+
+    public function station() : BelongsTo
+    {
+        return $this->belongsTo(Station::class);
+    }
+
+    public function trainingPlaces() : HasMany
+    {
+        return $this->hasMany(TrainingPlace::class);
+    }
+
+    public function waitingList() : BelongsTo
+    {
+        return $this->belongsTo(WaitingList::class);
+    }
+
+    /**
+    * Generate a list of available training positions for a waiting list
+    * based upon the places configured and also related
+    *
+    * @return array
+    */
+    public static function availablePlacesForWaitingList(WaitingList $waitingList)
+    {
+        $positions = self::with('station')
+            ->withCount('trainingPlaces')
+            ->where('waiting_list_id', $waitingList->id)
+            ->get();
+
+        return $positions->reject(function ($position) {
+            return (int)$position->training_places_count === (int)$position->places;
+        })->map(function ($position) {
+            return ['id' => $position->id, 'callsign' => $position->station->callsign];
+        })->toArray();
+    }
+}
--- a/app/Notifications/Training/TrainingPlaceOffer.php
+++ b/app/Notifications/Training/TrainingPlaceOffer.php
@@ -6,6 +6,7 @@ use Illuminate\Bus\Queueable;
 use Illuminate\Notifications\Messages\MailMessage;
 use Illuminate\Notifications\Notification;
 
+// TODO: Flesh out notification
 class TrainingPlaceOffer extends Notification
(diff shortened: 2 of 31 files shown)


app/Models/Training/TrainingPlace/TrainingPosition.php
```diff
@@ -0,0 +1,51 @@
...
+        $positions = self::with('station')
+            ->withCount('trainingPlaces')
+            ->where('waiting_list_id', $waitingList->id)
+            ->get();
+
+        return $positions->reject(function ($position) {
+            return (int) $position->training_places_count === (int) $position->places;
+        })->map(function ($position) {
+            return ['id' => $position->id, 'callsign' => $position->station->callsign];
+        })->toArray();
```
CalumTowers, line 49: Could you just chain this to get rid of the temp var?

app/Notifications/Training/TrainingPlaceOffer.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+namespace App\Notifications\Training;
+
+use Illuminate\Bus\Queueable;
+use Illuminate\Notifications\Messages\MailMessage;
+use Illuminate\Notifications\Notification;
+
+class TrainingPlaceOffer extends Notification
```
CalumTowers, line 9: Might be worth putting a TODO in this class just so its super obvious it isn't ready

database/migrations/2021_08_28_100948_create_training_place_offers_table.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('training_place_offers', function (Blueprint $table) {
+            $table->uuid('offer_id');
```
CalumTowers, line 17: I'm not sure how I feel about this - introducing a mix of tables with uuids and tables with ids. My gut feeling is that we should stick to the convention we have in place already of ids and then use the same method as V/T to generate an offer id (see application references for info)

tests/Feature/Training/TrainingPlace/TrainingPlaceOfferUserTest.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Tests\Feature\Training\TrainingPlace;
+
+use Tests\TestCase;
+
+class TrainingPlaceOfferUserTest extends TestCase
```
CalumTowers, line 7: TODO

CalumTowers (commented):
Seems to be going the right direction, nice work! 👍

app/Models/Training/TrainingPlace/TrainingPosition.php
AxonC: I thought it was a little more readable like this. Lots of faff to read otherwise

database/migrations/2021_08_28_100948_create_training_place_offers_table.php
AxonC: I did think about it. I think my overriding motivation was not to have a public-facing auto-incrementing ID. Especially if we want to move this to nexus in the future, where we presumably might use uuids more?

tests/Feature/Training/TrainingPlace/TrainingPlaceOfferUserTest.php
AxonC: done

app/Notifications/Training/TrainingPlaceOffer.php
AxonC: done

PR #2398, opened 2022-01-30
rep: Modernise Mail Config

atoff:
NB: Will likely require changing the production environment settings


CalumTowers (approved):
Env updated on prod

PR #2454, opened 2022-02-20
feat: Remove IPB files dependecy

atoff:
Replaces use of locally installed IPB instance core files with database calls to remove dependency for having IPS installed on same machine as core for forum syncing.


CalumTowers (approved):
Spot on @atoff, thank you!

PR #2670, opened 2022-08-19
refactor: clarify gatwick gnd role name

AxonC:


database/migrations/2022_08_19_183038_rename_gatwick_gnd_students_role.php
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
```
atoff, line 25: I think it might be good to include the down migration just in case
```suggestion
        DB::table('mship_role')->where(['name' => 'Gatwick GND Students'])->update(['name' => 'Gatwick Students']);
```

atoff (approved):

PR #2671, opened 2022-08-19
feat: stop non-division members from being added to waiting lists

AxonC:


app/Nova/Actions/Training/AddStudentToWaitingList.php
```diff
@@ -27,7 +27,7 @@ class AddStudentToWaitingList extends Action
      * @param  \Illuminate\Support\Collection  $models
      * @return mixed
      */
-    public function handle(ActionFields $fields, Collection $models)
+    public function handle(ActionFields $fields, Collection $models) // NOSONAR
```
atoff, line 30: Why no sonar?

AxonC: Doesn't like the number of return types

atoff (approved):

PR #2672, opened 2022-08-20
feat: add job to remove non-home members from lists

AxonC:
Trello: 233

This establishes logic to remove any existing members from the list where they are no longer members of the UK division. They will be directly removed from the waiting list and then an email sent directly to the user. 

Few things to note:
- This relies on their state being updated either via a login or via the discovery scripts (which could be a maximum of 7 days lag time depending on where they have come in the import scripts)
- I decided to do this via queues just because this scales much more efficiently for the larger waiting lists, even if the messages will for the most part just be discarded from the queue.
- A command will run once daily at 5am to check all ATC waiting lists

Complements #2671

### DIFF
--- a/app/Jobs/Training/CheckHomeMemberInWaitingList.php
+++ b/app/Jobs/Training/CheckHomeMemberInWaitingList.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace App\Jobs\Training;
+
+use App\Models\Mship\Account;
+use App\Models\Training\WaitingList;
+use App\Notifications\Training\RemovedFromWaitingListNonHomeMember;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\Log;
+
+class CheckHomeMemberInWaitingList implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    private WaitingList $waitingList;
+    private Account $account;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(WaitingList $waitingList, Account $account)
+    {
+        $this->waitingList = $waitingList;
+        $this->account = $account;
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        try {
+            $this->waitingList->accounts()->find($this->account->id);
+        } catch (ModelNotFoundException) {
+            Log::warning("Account {$this->account->id} not in waiting list.");
+            return;
+        }
+
+        if (!$this->account->primary_state->isDivision) {
+            $this->waitingList->removeFromWaitingList($this->account);
+            $this->account->notify(new RemovedFromWaitingListNonHomeMember);
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


app/Jobs/Training/CheckHomeMemberInWaitingList.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        try {
+            $this->waitingList->accounts()->find($this->account->id);
```
atoff, line 42: Will this throw? Thought it is only `firstOrFail`

resources/views/emails/training/waiting_list_non_home_removal.blade.php
```diff
@@ -0,0 +1,9 @@
+@extends('emails.messages.post')
+
+@section('body')
+
+<p>As you have changed your VATSIM Division away from the UK Division you are no longer eligible to receive ATC training in the UK. You were on one of the waiting lists for ATC training in the UK and as such as you are no longer a member of the UK Division you have been removed from the waiting list.</p>
```
atoff, line 5:
```suggestion
<p>As you have changed your VATSIM Division away from the UK Division you are no longer eligible to receive ATC training in the UK. You were on one of the waiting lists for ATC training in the UK and, as you are no longer a member of the UK Division, you have been removed from this waiting list.</p>
```

tests/Unit/Training/WaitingList/CheckHomeMemberscCommandTest.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Tests\Unit\Training\WaitingList;
+
+use App\Jobs\Training\CheckHomeMemberInWaitingList;
+use App\Models\Mship\Account;
+use App\Models\Training\WaitingList;
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+use Illuminate\Support\Facades\Queue;
+use Tests\TestCase;
+
+class CheckHomeMemberscCommandTest extends TestCase
```
atoff, line 12:
```suggestion
class CheckHomeMembersCommandTest extends TestCase
```

resources/views/emails/training/waiting_list_non_home_removal.blade.php
AxonC: That'll teach me from copying and pasting from the trello ticket :D

CalumTowers:
Appreciate this is a bit of a rework but would a LeftUKDivisionEvent be more efficient here? Fire it when their state changes and remove them from the waiting list off the back of it?

app/Jobs/Training/CheckHomeMemberInWaitingList.php
AxonC: You can tell I was undecided about this - forgot to change it back. Sorted now

tests/Unit/Training/WaitingList/CheckHomeMemberscCommandTest.php
AxonC: Amended and renamed file

AxonC:
> Appreciate this is a bit of a rework but would a LeftUKDivisionEvent be more efficient here? Fire it when their state changes and remove them from the waiting list off the back of it?

Possibly. Suffers from the same issue with the update scripts so I wouldn't say it provides a massive jump in efficiency. Would be worth refactoring I think if we ever get the webhooks from .NET?

AxonC:
> Appreciate this is a bit of a rework but would a LeftUKDivisionEvent be more efficient here? Fire it when their state changes and remove them from the waiting list off the back of it?

FWIW I had considered responding to the AccountAltered event but that fired on quite a number of conditions so decided against it for this purpose.

CalumTowers:
> > Appreciate this is a bit of a rework but would a LeftUKDivisionEvent be more efficient here? Fire it when their state changes and remove them from the waiting list off the back of it?
> 
> Possibly. Suffers from the same issue with the update scripts so I wouldn't say it provides a massive jump in efficiency. Would be worth refactoring I think if we ever get the webhooks from .NET?

LeftUkDivision would only fire when someone leaves the UK division though? And it would be part of an existing process and avoid the need for us to run this sort of command regularly?

CalumTowers:
@AxonC Can we expand this to also kick people off a waiting list if they become inactive?

PR #2798, opened 2022-11-21
feat: Waiting list automated removals

BenEsen97:
This PR adds functionality to have waiting list members automatically removed as directed in the ATC training handbook, when their network activity falls below the requirements.

The work adds a scheduled task which runs and performs the following functions:

- Identifies members of a waiting list who do not meet the activity requirements and are not already marked for removal, marking them for removal
- Checks waiting list members who are marked for removal and actions the removal process (sending a 5 day reminder, and finally removing them)
- Cancels pending removal action when members meet the activity requirement
- Emails are sent out throughout this process, and the relevant notifications and email views have been created.

Finally, the UI has been updated for viewing waiting lists to provide members with information relating to this process.

This is my first contribution to core, and I am happy to receive feedback and make any required amendments.

**This PR relates to an upcoming training policy change, and should not be merged without prior co-ordination with the ATC training director.**


database/migrations/2022_11_06_111641_add_training_list_enforce_hour_eligibility_column.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('training_waiting_list', function (Blueprint $table) {
+            $table->enum('enforce_hour_requirement', ['yes', 'no'])->after('flags_check')->defalt('yes');
```
AxonC, line 17: Why use an enum here when it is a true/false value?

app/Notifications/Training/WaitingListRemovalAdded.php
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Get the notification's delivery channels.
+     *
+     * @param  mixed  $notifiable
+     * @return array
+     */
+    public function via($notifiable)
+    {
+        return ['mail', 'database'];
```
AxonC, line 39:
```suggestion
        return ['mail'];
```
You haven't defined an array representation, so should be via mail.

app/Notifications/Training/WaitingListRemovalAdded.php
```diff
@@ -0,0 +1,68 @@
...
+
+    private $list_name;
+    private $removal_date;
+
+    /**
+     * Create a new notification instance.
+     *
+     * @return void
+     */
+    public function __construct(String $list_name, Carbon $removal_date)
```
AxonC, line 23:
```suggestion
    public function __construct(string $list_name, Carbon $removal_date)
```

app/Notifications/Training/WaitingListRemovalAdded.php
```diff
@@ -0,0 +1,68 @@
...
+use Carbon\Carbon;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Notifications\Messages\MailMessage;
+
+class WaitingListRemovalAdded extends Notification implements ShouldQueue
+{
+    use Queueable;
+
+    private $list_name;
```
AxonC, line 15: Since we are using PHP 8, I'd suggest type-hinting these class variables now

REF: https://www.phpclasses.org/blog/post/1047-php-8-type-hinting.html#th_types_prop

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,160 @@
+<?php
```
AxonC, line 1: I cannot find any tests for any of the functions in this file.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,160 @@
...
+     * @var string
+     */
+    protected $description = 'Checks ATC hour totals for members who are on an ATC waiting list and marks inactive members for automatic removal from the list.';
+
+    /**
+     * A local cache of account activity to prevent duplicate calculations of a members hours
+     * 
+     * @var array
+     */
+    protected $accountActivityCache = [];
```
AxonC, line 39: I don't see the value in storing this as a class-level variable.

You'd make your function `processWaitingListAccountsMarkedForRemoval` much more testable if you could make it more 'pure' by passing in the list of accounts you think should be marked for removal.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,160 @@
...
+       }
+       
+    }
+
+    /**
+     * Retrieves training waiting lists which enforce hour requirements
+     */
+    protected function retrieveHourEnforcedWaitingLists()
+    {
+        return WaitingList::where('enforce_hour_requirement', '=', 'yes')->get();
```
AxonC, line 58: This is a good example where a string enum isn't always the nicest. A simple true/false comparison here would simplify the code.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,160 @@
...
+     */
+    protected function identifyNewNonEligibleWaitingListAccounts(WaitingList $waitingList)
+    {
+        // We only want to find Active waiting list accounts which are not already marked for removal
+        $waitingListAccounts                                        = $waitingList->accounts
+            ->filter(function($account){
+                return (
+                    $account->pivot->current_status->name == 'Active' &&
+                    (
+                        is_null($account->pivot->pending_removal->removal_date ?? null) ||
```
AxonC, line 74:
```suggestion
                        !!$account->pivot->pending_removal->removal_date ||
```
A null value evaluates to falsy, so the additional logic is unnecessary.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,160 @@
...
+     * Performs two actions:
+     * 
+     * - Checks to see if members who are pending removal have now met the hour criteria
+     * - Sends reminder emails as per policy schedule, and actions waiting list removal
+     * 
+     * @param WaitingList $waitingList
+     */
+    protected function processWaitingListAccountsMarkedForRemoval(WaitingList $waitingList)
+    {
+         $waitingListAccounts                                        = $waitingList->accounts
```
AxonC, line 113: This is what I mean - here, the logic is duplicated, where you could have just passed in a list of accounts you want to start the removal process for.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,160 @@
...
+                    )
+                );
+            });
+        
+        foreach($waitingListAccounts as $account){
+            // Calculate the total activity within the last 3 months
+            if(isset($this->accountActivityCache[$account->id])){
+                $activeMinutes                                      = $this->accountActivityCache[$account->id];
+            } else {
+                $activeMinutes                                      = Atc::where('account_id', $account->id)
```
AxonC, line 85: Regardless of my recommendation on removing the local cache, this logic is duplicated from https://github.com/VATSIM-UK/core/blob/c3ac806defa080dd5a67be737aae3c24788148e1/app/Models/Training/WaitingList/WaitingListAccount.php#L122

Since you are already iterating over instances of `WaitingListAccount` you should just call this method.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,160 @@
...
+            $this->identifyNewNonEligibleWaitingListAccounts($waitingList);
+            $this->processWaitingListAccountsMarkedForRemoval($waitingList);
+       }
+       
+    }
+
+    /**
+     * Retrieves training waiting lists which enforce hour requirements
+     */
+    protected function retrieveHourEnforcedWaitingLists()
```
AxonC, line 56: Every ATC waiting list, by virtue of being that type, has an hour check built in by default (see : https://github.com/VATSIM-UK/core/blob/c3ac806defa080dd5a67be737aae3c24788148e1/app/Models/Training/WaitingList/WaitingListAccount.php#L138)

I'd argue that actually this column is not required on this basis and that you should just filter for ATC list types.

AxonC (changes requested):
General Comments:

Given the scale of the number of members on lists, I'd suggest that some of the jobs, like marking for approval, detect if those marked for approval are now back into eligibility, and sending the notifications are done via an event-driven architecture. In Laravel terms, this is where an event is emitted and a queued listener responds to the relevant action. Blocking the command until the completion of this process could lead to performance issues.

This would make it easier for other things in the future to start the process of marking them for removal e.g. if it was that we wanted to start looking for activity on specific positions, or whether a theory exam had been completed within X time of joining the list, for example. You'd be able to trigger from multiple sources, rather than coupling to the implementation of the command.

You may wish to have a separate command that runs more frequently to check the removals so you are avoiding the edge conditions you have with the daily checks. Separating the command out would be my advice in general here.

database/migrations/2022_11_06_111641_add_training_list_enforce_hour_eligibility_column.php
BenEsen97: This mainly stems from personal preference as opposed to any benefit in this specific situation. 

My way of thinking is generally if we use 1/0 for this column, and in the future there is a case for expanding this to something along the lines of:

Enforces Hour Requirement
- yes
- reduced
- no

Then there are minimal database updates required. The above may not be the best example, and I wouldn't argue against it being a Boolean for this column, but the reasoning is to provide some flexibility / scalability

app/Notifications/Training/WaitingListRemovalAdded.php
BenEsen97: Thanks Callum, I wasn't entirely sure what the purpose of this was and aired on the side of caution when referencing similar classes. I will change this.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
BenEsen97: Currently it is the case that all ATC waiting lists require the hour check, however, when discussing Adam's plans with him to see what I could assist with, he expressed interest in migrating the OBS-S1 waiting list from CTS to Core. 

In that event, the waiting list would not require an hour check, so the hard coded solution currently in place would need changing. 

Since I was working heavily on the hour check aspect of the waiting list in this task, it made sense to add this as a waiting list specific setting now.

AxonC: As you say, many more changes are required to have an ATC list that doesn't enforce any hour requirements, so I'd suggest keeping the status quo until the need arises.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
BenEsen97: If there is not a pending removal row for this waiting list account, then you will be trying to access removal_date on null. The same could be achieved by doing $account->pivot->pending_removal?->removal_date resulting in slightly less code than the current implementation.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
BenEsen97: > Regardless of my recommendation on removing the local cache, this logic is duplicated from https://github.com/VATSIM-UK/core/blob/c3ac806defa080dd5a67be737aae3c24788148e1/app/Models/Training/WaitingList/WaitingListAccount.php#L122
> 
> 
> 
> Since you are already iterating over instances of `WaitingListAccount` you should just call this method. 

I have just realised why I did it this way, and it's because the existing method caches with a 24hour TTL.

That is (in my opinion) an unacceptable amount of caching for this particular function, and if a user meets eligibility after being told that they are on the removal process, then they should be informed that this is acknowledged by the system as soon as it reasonably practical (in this case it will be a maximum of 12hours if the scheduled task runs twice daily)

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
BenEsen97: Hi @AxonC 

This function accepts a waiting list, and then discovers accounts which are marked for removal for that waiting list.

I don't fully understand how the logic is duplicated, as each time the function is called, the waiting list will be different - so line 113 retrieves the relevant accounts for the current waiting list?

BenEsen97:
> General Comments:
> 
> Given the scale of the number of members on lists, I'd suggest that some of the jobs, like marking for approval, detect if those marked for approval are now back into eligibility, and sending the notifications are done via an event-driven architecture. In Laravel terms, this is where an event is emitted and a queued listener responds to the relevant action. Blocking the command until the completion of this process could lead to performance issues.
> 
> This would make it easier for other things in the future to start the process of marking them for removal e.g. if it was that we wanted to start looking for activity on specific positions, or whether a theory exam had been completed within X time of joining the list, for example. You'd be able to trigger from multiple sources, rather than coupling to the implementation of the command.
> 
> You may wish to have a separate command that runs more frequently to check the removals so you are avoiding the edge conditions you have with the daily checks. Separating the command out would be my advice in general here.
> 
> I'll do a second review if you have any questions. Feel free to reach out on DIscord with any other questions.

Hi @AxonC 

I have refactored this PR, including splitting the system into two isolated commands that be be ran on separate schedules.

The first command is responsible for checking waiting list accounts meet the activity requirement, both those not marked for removal and those which are. 

The second command is responsible for actioning pending removals.

I've also created unit tests to cover both of the new commands.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,63 @@
...
+     * @var string
+     */
+    protected $description = 'Checks ATC waiting list activity rules are being met for each account';
+
+    /**
+     * Executes all necessary console commands.
+     */
+    public function handle()
+    {
+        WaitingList::all()->filter(function ($waitingList) {
```
AxonC, line 35: You're going to have an N+1 query problem here. You are accessing the accounts of a waiting list 'lazily,' which means you are doing more queries than necessary (see: https://laravel.com/docs/9.x/eloquent-relationships#eager-loading)

app/Console/Commands/WaitingLists/ProcessPendingRemovals.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    protected $description = 'Processing waiting list accounts marked for automatic removal';
+
+    /**
+     * Executes all necessary console commands.
+     */
+    public function handle()
+    {
+        WaitingList::all()->filter(function ($waitingList) {
+            return $waitingList->isAtcList();
```
AxonC, line 33: I'd suggest since you're using this filter for ATC lists in more than one place,  you create a local scope: https://laravel.com/docs/9.x/eloquent#local-scopes

database/migrations/2022_11_06_201538_create_training_waiting_list_account_pending_removal_table.php
```diff
@@ -0,0 +1,37 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('training_waiting_list_account_pending_removal', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedInteger('waiting_list_account_id');
+            $table->enum('status', ['Pending', 'Completed', 'Cancelled'])->default('Pending');
```
AxonC, line 19: The enum, for me, isn't serving any purpose since you're not tracking the changes between the statuses.

I think you'd be better off having columns like:

`pending_at`
`completed_at`
`cancelled_at` 

So you at least then have an audit trail. For querying those which are pending but not yet completed, you would get those where pending_at is not null, but completed at was (and further cancelled_at is null too).

database/migrations/2022_11_06_201538_create_training_waiting_list_account_pending_removal_table.php
```diff
@@ -0,0 +1,37 @@
...
+            $table->unsignedInteger('emails_sent')->default(0);
+            $table->timestamp('removal_date');
```
AxonC, line 21:
```suggestion
            $table->timestamp('removed_at');
```
This isn't just a date but a datetime and should be named accordingly.

resources/views/emails/training/waiting_list_removal_added.blade.php
```diff
@@ -0,0 +1,23 @@
+@extends('emails.messages.post')
```
AxonC, line 1: Nit:  Missing line endings on all the email blade files.

database/migrations/2022_11_06_201538_create_training_waiting_list_account_pending_removal_table.php
AxonC: You could then convert these into methods on the model to save duplication and avoid having to reference the raw enum value each time. You then abstract away the underlying implementation leaving it more flexible in the future should you, say, want to change what constitutes pending

resources/views/mship/waiting-lists/view.blade.php
```diff
@@ -36,6 +36,14 @@
                             @endif</td>
                     </tr>
                 </table>
+                @if(
+                    $list->pivot->current_status == 'Active' &&
```
AxonC, line 40: This is a great example of where my column based approach for tracking times and model methods would work very well.

app/Notifications/Training/WaitingListRemovalCancelled.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toMail($notifiable)
+    {
+        $subject = 'You now meet the eligibility for a waiting list you are a member of';
```
AxonC, line 47: Why not specify the list(s) given you specify it in the first email sent for waiting list removal added?

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
AxonC: That would mean the email would have a different timespan for the people viewing the waiting lists. This I find to be unacceptable. They need to use the same method of detecting activity, or if someone raises an issue, the human who does the follow-up gets a different view of the system doing the automated checking.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,160 @@
...
+            } else {
+                $activeMinutes                                      = Atc::where('account_id', $account->id)
+                    ->whereDate('disconnected_at', '>=', Carbon::parse('3 months ago'))
+                    ->isUk()
+                    ->sum('minutes_online');
+                $this->accountActivityCache[$account->id]           = $activeMinutes;
+            }
+
+            // When the hours are less than the requirement, set a removal date and notify the member
+            if(((int)$activeMinutes ?? 0) < $this->minutesRequired){
```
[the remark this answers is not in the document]
AxonC: Why was this resolved?

app/Console/Kernel.php
```diff
@@ -59,6 +59,10 @@ protected function schedule(Schedule $schedule)
             ->hourlyAt(15)
             ->graceTimeInMinutes(15);
 
+        $schedule->command('waitinglists:processpendingremovals')
```
AxonC, line 62: I don't think this interval is going to work. There is no assurance on the processing time of incoming network data from the datafeed. I think once daily is the best assurance you will get.

app/Listeners/Training/WaitingList/SendAccountRegainedActivityNotification.php
```diff
@@ -0,0 +1,30 @@
...
+use App\Notifications\Training\WaitingListRemovalCancelled;
+
+class SendAccountRegainedActivityNotification
+{
+    /**
+     * Create the event listener.
+     *
+     * @return void
+     */
+    public function __construct()
```
AxonC, line 15: Remove empty constructor.

app/Listeners/Training/WaitingList/SendAccountRemovedDueToActivityNotification.php
```diff
@@ -0,0 +1,30 @@
...
+use App\Notifications\Training\WaitingListRemovalCompleted;
+
+class SendAccountRemovedDueToActivityNotification
+{
+    /**
+     * Create the event listener.
+     *
+     * @return void
+     */
+    public function __construct()
```
AxonC, line 15: Remove empty constructor.

app/Listeners/Training/WaitingList/SendWaitingListRemovalFiveDayReminderNotification.php
```diff
@@ -0,0 +1,30 @@
...
+use App\Notifications\Training\WaitingListRemovalReminder;
+
+class SendWaitingListRemovalFiveDayReminderNotification
+{
+    /**
+     * Create the event listener.
+     *
+     * @return void
+     */
+    public function __construct()
```
AxonC, line 15: Remove empty constructor.

app/Notifications/Training/WaitingListRemovalCancelled.php
```diff
@@ -0,0 +1,54 @@
...
+     * Create a new notification instance.
+     *
+     * @return void
+     */
+    public function __construct(string $list_name)
+    {
+        parent::__construct();
+
+        $this->list_name = $list_name;
+    }
```
AxonC, line 26:
```suggestion
    /**
     * Create a new notification instance.
     *
     * @return void
     */
    public function __construct(private string $list_name)  { parent::__construct(); }
```

If you need to do no further work to the inbound properties of the constructor, consider using PHP8's property promotion: https://stitcher.io/blog/constructor-promotion-in-php-8

AxonC: I'd extend this to any class created in this PR where the constructor mainly just assigns some class level variables.

app/Console/Commands/WaitingLists/ProcessPendingRemovals.php
```diff
@@ -0,0 +1,59 @@
...
+        })->each(function (WaitingList $waitingList) {
+            $waitingList->accounts
```
AxonC, line 35: N+1 problem here also.

app/Console/Commands/WaitingLists/ProcessPendingRemovals.php
```diff
@@ -0,0 +1,59 @@
...
+            ->filter(function ($account) {
+                return $account->pivot->current_status->name == 'Active';
+            })->filter(function ($account) {
```
AxonC, line 38: Another example of where this logic should be abstracted to a model method.

app/Console/Commands/WaitingLists/ProcessPendingRemovals.php
```diff
@@ -0,0 +1,59 @@
...
+                return $account->pivot->pending_removal?->status == 'Pending';
+            })->each(function ($account) use ($waitingList) {
+                // Send 5 day reminders
+                if (
+                    Carbon::parse($account->pivot->pending_removal->removal_date)->subDays(6) <= Carbon::now() &&
```
AxonC, line 43: The magic number 6 has no relationship to the comment - please explain why in a comment subDays(6) is used since it isn't self-explanatory.

app/Listeners/Training/WaitingList/SendAccountMarkedForRemovalNotification.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Create the event listener.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        //
+    }
```
AxonC, line 18:
```suggestion
```

tests/Unit/Training/WaitingList/WaitingListCheckAccountMeetActivityRequirementTest.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        parent::setUp();
+
+        $this->waitingList = factory(WaitingList::class)->create();
+    }
+
+    /** @test */
+    public function itShouldMarkInactiveAccountForRemoval()
+    {
+        $account = factory(Account::class)->create()->refresh();
```
AxonC, line 28: Explain why the refresh is needed on a newly created model via comment.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,63 @@
...
+        WaitingList::all()->filter(function ($waitingList) {
+            return $waitingList->isAtcList();
+        })->each(function (WaitingList $waitingList) {
+            $waitingList->accounts
+            ->filter(function ($account) {
+                return $account->pivot->current_status->name == 'Active';
+            })->each(function ($account) use ($waitingList) {
+                $activeMinutes = Atc::where('account_id', $account->id)
+                    ->whereDate('disconnected_at', '>=', Carbon::parse('3 months ago'))
+                    ->isUk()
```
AxonC, line 44:
```suggestion
                    ->isUk()
                    ->whereDate('disconnected_at', '>=', Carbon::parse('3 months ago'))
```

Filtering down the amount of rows you need to filter on `disconnected_at` since I don't recall there being an index on that access pattern of columns.

AxonC (changes requested):

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
```diff
@@ -0,0 +1,63 @@
...
+                    ->sum('minutes_online');
+
+                if ($activeMinutes < $this->minutesRequired) {
+                    if (is_null($account->pivot->pending_removal?->removal_date) || $account->pivot->pending_removal->status != 'Pending') {
+                        $removalDate = Carbon::now();
+                        $removalDate->addDays(31); // When calculating date diff, this will be a difference of 30 days
```
AxonC, line 50: I'd suggest rather than relying on this, a consistent processing time should be used, or there needs to be more edge case testing of this.

app/Console/Commands/WaitingLists/ProcessPendingRemovals.php
BenEsen97: Hi @AxonC 

I have created a local scope for this exact purpose in https://github.com/VATSIM-UK/core/pull/2804 as their was a similar requirement in that PR, and a local scope was suggested by @atoff.

Makes sense to utilize that here, however this change is blocked until 2804 is merged.

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
BenEsen97: Hi @AxonC 

That is fair, however it is worth considering if this caching can be reduced - perhaps halved to 12 hours? 

What would the performance cost be of this?

database/migrations/2022_11_06_201538_create_training_waiting_list_account_pending_removal_table.php
BenEsen97: I have renamed this to `remove_at `because it is a flag for when an account should be removed from a list (in the future).

app/Console/Commands/WaitingLists/CheckMembersMeetActivityRules.php
BenEsen97: Marking this as resolved now to clean up the PR, as this will be resolved by the scope change (https://github.com/VATSIM-UK/core/pull/2798#discussion_r1046254036) when the blocking PR is merged.

app/Console/Commands/WaitingLists/ProcessPendingRemovals.php
BenEsen97: Hi @AxonC 

I am struggling with this one a little - can you provide an example of the logic?

I don't work with Laravel in my day-to-day so lack an in-depth knowledge of it at this level. 

The main issue I am facing is that as far as I can see, to add a model method such as `activeAccountsPendingRemoval()` to replace the following two filters:

```
->filter(function ($account) {
    return $account->pivot->current_status->name == 'Active';
})->filter(function ($account) {
    return $account->pivot->pending_removal?->status == 'Pending';
})
```

I would need to add it to the `Account` model, not the `WaitingListAccount` model. Adding it to the `WaitingListAccount` seems like the correct place, but is then not accessible in the query builder as far as I can see due to it being on a pivot and not the parent model.

I believe I could do something like:

```
->filter(function ($account) {
    return $account->pivot->activeAccountsPendingRemoval();
})
```

But using a filter appears to defeat the point of the change.

PR #2799, opened 2022-11-23
feat: remove non division members & inactive members no state change

AxonC:
Adds listeners to `AccountAltered` event, which, when triggered, will remove:

- non-division members from waiting lists they are a part of on the first detection of a state other than Division
- accounts marked as 'inactive', aka their rating becoming `-1` internally.

When the removal happens, they will be sent an email depending on which condition they have no longer met to be on the waiting list.


app/Notifications/Training/RemovedFromWaitingListInactiveAccount.php
```diff
@@ -0,0 +1,37 @@
...
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toMail($notifiable)
+    {
+        return (new MailMessage)
+        ->from('support@vatsim.uk', 'VATSIM UK - Member Services')
+        ->subject('UK ATC training waiting list removal')
```
atoff, line 34: I believe we have pilot training lists too, perhaps this should be
```suggestion
        ->subject('UK training waiting list removal')
```

app/Notifications/Training/RemovedFromWaitingListNonHomeMember.php
```diff
@@ -0,0 +1,37 @@
...
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toMail($notifiable)
+    {
+        return (new MailMessage)
+                    ->from('support@vatsim.uk', 'VATSIM UK - Member Services')
+                    ->subject('UK ATC training waiting list removal')
```
atoff, line 34: As above

app/Notifications/Training/RemovedFromWaitingListInactiveAccount.php
CalumTowers: Title Case Feels Appropriate For A Subject Too

app/Notifications/Training/RemovedFromWaitingListNonHomeMember.php
CalumTowers: Title Case Feels Appropriate For A Subject Too

CalumTowers (commented):
Can we just double check the consistency on the debug logs between the two listeners (e.g. the text that is logged and at one stage).

Otherwise I'm happy.

app/Notifications/Training/RemovedFromWaitingListNonHomeMember.php
AxonC: Thanks, amended.

app/Notifications/Training/RemovedFromWaitingListInactiveAccount.php
AxonC: Amended as suggested.

AxonC:
> Can we just double check the consistency on the debug logs between the two listeners (e.g. the text that is logged and at one stage).
> 
> Otherwise I'm happy.

Good shout. Amended to have the best level of detail across the board.

resources/views/emails/training/waiting_list_non_home_removal.blade.php
```diff
@@ -0,0 +1,9 @@
+@extends('emails.messages.post')
+
+@section('body')
+
+<p>As you have changed your VATSIM Division away from the UK Division you are no longer eligible to receive training in the UK. You were on one of the waiting lists for ATC training in the UK and, as you are no longer a member of the UK Division, you have been removed from this waiting list.</p>
```
atoff, line 5:
```suggestion
<p>As you have changed your VATSIM Division away from the UK Division you are no longer eligible to receive training in the UK. You were on one of the waiting lists for training in the UK and, as you are no longer a member of the UK Division, you have been removed from this waiting list.</p>
```

atoff (approved):

Issue #10, opened 2016-12-03
Convert all classes to PSR-4 namespacing (Old 343)


NFarrington:
In gitlab by @A-Lawrence on Aug 31, 2015, 01:07

null

PR #2804, opened 2022-11-25
feat: waiting list top ten notifications

BenEsen97:
This PR adds functionality to the waiting list system to send a notification to members of ATC waiting lists when they reach position #10 or greater.


app/Listeners/Training/WaitingList/SendTopTenAtcNotifications.php
```diff
@@ -0,0 +1,52 @@
...
+                $matched = $matched + 1;
+                if ($account->pivot->top_ten_notified == 'no') { // We do not want to notify people who have already been notified of their top 10 position
+                    return true;
+                }
+
+                return false;
+            }
+
+            return false;
+        });
```
CalumTowers, line 44: `$event->waitingList->accountsByEligibility(true)->take(10);` should give you the top 10 of the waiting list selected. Then you can just filter out those that have not had the notification.

app/Notifications/Training/WaitingListAtcTopTen.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toMail($notifiable)
+    {
+        $subject = 'You are now top 10 in an ATC waiting list';
```
CalumTowers, line 47: This doesn't feel very description, and can be in-lined into the return statement itself.

resources/views/emails/training/waiting_list_atc_top_ten.blade.php
```diff
@@ -0,0 +1,25 @@
...
+<pre>
+        If you reach the top of the waiting list and a training place becomes available, but you have not yet passed the relevant CTS theory exam, you will be contacted by a member of the ATC Training staff. You will be given 7 days to attempt the theory exam. If unsuccessful, there is a 7 day cool down period, after which you will be given a further 8 days to complete the exam. This will continue until you pass the exam.
+
+        Should you not attempt the exam within the 7 days, you will be removed from the waiting list.
+</pre>
+
+<p>
+    If you have any issues with this and are still interested in training then reach out to us via the helpdesk with help topic 'Member Services'.
+</p>
+@stop
```
CalumTowers, line 25: I feel like this could do with some rewording. This should be an exciting email for our members to recieve to let them know that they'll soon be entering the world of training.

CalumTowers (changes requested):
I don't feel like an event, triggered by someone's removal from a list is a great approach to triggering this particular notification. An on from a waiting list isn't inherently tied directly to an order of that waiting list. For example, if someone else's date on the list changes, that would impact their position in the list.

Ordering is a bit tricky, as the position of an account is determined by eligibility and by the date they were added. It is possible for someone to be position 11, then 10, then 12 all within the same day based on other people's activity.

For that reason I would suggest instead putting this on a scheduled command where once per day, we pull the top 10 people on any given list, filter those that have had the notification and then work out whether we should send them a notification.

resources/views/emails/training/waiting_list_atc_top_ten.blade.php
BenEsen97: This email content was provided by Adam, my only contribution here was correcting a typo that I spotted. 

Content writing certainly isn't a strong point of mine, so I'll stay well away and let Adam know about this feedback. 😀

app/Listeners/Training/WaitingList/SendTopTenAtcNotifications.php
BenEsen97: That's very useful, cheers

app/Notifications/Training/WaitingListAtcTopTen.php
BenEsen97: Again, will await feedback from Adam on this

resources/views/emails/training/waiting_list_atc_top_ten.blade.php
1adamf: To be honest I didn't really think of it that way however I feel like exciting people when they reach top 10 could be counter-productive as 10th to 1st could be a matter of weeks or it could be a matter of months. The reason that we picked top 10 was to keep it conservative to ensure that everyone get plenty of notice to start working on the theoretical material.

I'll reword a little however it'll remain fairly similar to prevent people asking too many questions along the lines of "you said I'd have a training place soon and it's been 3 weeks"

CalumTowers: @1adamf I think my concern is that this is the first time some members may hear from us after 6-12 months and a lot of it centres around policy and how they could be removed. I know this is important information to convey but by making it a bit more "exciting" as I put it, may enthuse them to ensure they engage with the policy.

app/Console/Commands/WaitingLists/SendAtcTopTenNotification.php
```diff
@@ -0,0 +1,47 @@
...
+                return is_null($account->pivot->within_top_ten_notification_sent_at);
+            });
+
+            foreach ($nonNotifiedTopTenAccounts as $account) {
+                $account->notify(new WaitingListAtcTopTen($waitingList->name));
+                $account->pivot->within_top_ten_notification_sent_at = Carbon::now();
+                $account->pivot->save();
+            }
+        }
+    }
```
CalumTowers, line 46: I would use collections here to avoid temporary variables, but that is a personal opinion - there is nothing functionally wrong with the code here.

Something like the below should work (although I've not tested it).

```php
        WaitingList::all()->filter(function (WaitingList $waitingList) {
            return $waitingList->isAtcList();
        })->mapWithKeys(function (WaitingList $waitingList) {
            return [$waitingList->name =>
                $waitingList
                    ->accountsByEligibility(true)
                    ->take(10)
                    ->filter(function ($account) {
                        return !!$account->pivot->within_top_ten_notification_sent_at;
                    })
            ];
        })->each(function (Collection $accounts, string $waitingListName) {
            $accounts->each(function ($account) use ($waitingListName) {
                $account->notify(new WaitingListAtcTopTen($waitingListName));
                $account->pivot->within_top_ten_notification_sent_at = Carbon::now();
                $account->pivot->save();
            });
        });
```

BenEsen97 (commented):
Feedback has been actioned, thanks for the review and I'm looking forward to reviewing any further comments.

app/Console/Commands/WaitingLists/SendAtcTopTenNotification.php
```diff
@@ -0,0 +1,50 @@
...
+    protected $description = 'Sends the top 10 members of an ATC waiting list a notification';
+
+    /**
+     * Executes all necessary console commands.
+     */
+    public function handle()
+    {
+        WaitingList::all()->filter(function ($waitingList) {
+            return $waitingList->isAtcList();
+        })->mapWithKeys(function (WaitingList $waitingList) {
```
atoff, line 34: This seems a little inefficient (we don't have many waiting lists, but hey, best practice 😉)
It would be better to implement the filter at the database level like so (you could possibly make this a scope filter - see https://laravel.com/docs/9.x/eloquent#local-scopes, but I will leave this up to you)
```suggestion
        WaitingList::where('department', WaitingList::ATC_DEPARTMENT)->get()->mapWithKeys(function (WaitingList $waitingList) {
```

app/Notifications/Training/WaitingListAtcTopTen.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toMail($notifiable)
+    {
+        return (new MailMessage)
+            ->from('atc-team@vatsim.uk', 'VATSIM UK - ATC Training')
```
atoff, line 48:
```suggestion
            ->from('atc-training@vatsim.uk', 'VATSIM UK - ATC Training')
```

app/Notifications/Training/WaitingListAtcTopTen.php
```diff
@@ -0,0 +1,52 @@
...
+            ->subject('You are now top 10 in an ATC waiting list')
```
atoff, line 49:
```suggestion
            ->subject("Within top 10 in {$this->list_name} waiting list")
```

atoff (changes requested):

app/Notifications/Training/WaitingListAtcTopTen.php
BenEsen97: I've updated this, but with a slight change as I feel the following reads better:

'Within top 10 on the <list> waiting list'

app/Console/Commands/WaitingLists/SendAtcTopTenNotification.php
BenEsen97: Hi @atoff 

Thank you for the review / feedback, much appreciated!

I've updated this to use a new ATC scope, which I believe makes the most sense in this context (and may prove to be useful in the future).

I completely agree that it's best to form good habits and scope the retrieval of data at the database end, for the best performance.

app/Models/Training/WaitingList.php
```diff
@@ -88,6 +89,17 @@ public function flags()
...
 
+    /**
+     * Scope a query to only include ATC waiting lists.
+     *
+     * @param  \Illuminate\Database\Eloquent\Builder  $query
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopeAtc($query)
+    {
+        return $query->where('department', '=', 'atc');
```
atoff, line 100: Seeing as we have this defined as a ENUM constant up top, we should use that for easier refactoring in future / single-source-of truth
```suggestion
        return $query->where('department', '=', self::ATC_DEPARTMENT);
```

atoff (changes requested):

atoff (approved):
Looks good to me Ben, thanks for this! I'll give @CalumTowers a change to make sure he is happy to and then we can merge!

CalumTowers (commented):
Can we just get some very high level tests added to confirm that people don't get the notification sent to them more than once?

New waiting list
Add 11 people to it
Run command
Assert 10 notifications sent
Run command
Assert no notifications sent
Remove top person from waiting list
Assert one notification sent

That should cover it.

BenEsen97:
> Can we just get some very high level tests added to confirm that people don't get the notification sent to them more than once?
> 
> New waiting list Add 11 people to it Run command Assert 10 notifications sent Run command Assert no notifications sent Remove top person from waiting list Assert one notification sent
> 
> That should cover it.

Hi @CalumTowers 

I have created a test to sanity check the new functionality, please review when you can.

tests/Unit/Training/WaitingList/WaitingListTest.php
```diff
@@ -173,4 +177,37 @@ public function itCanHaveFlagsRemoved()
...
+            });
+        });
+
+        Notification::fake();
+
+        $this->artisan('waitinglists:sendatctoptennotification');
+
+        Notification::assertTimesSent(10, WaitingListAtcTopTen::class);
+
+        Notification::fake();
```
CalumTowers, line 199: You should only need to call this once per test.

BenEsen97: Hi @CalumTowers 

From my testing, calling this resets the count of notifications, so without it being recalled, `assertNotSent` fails because the previous sub-test called the notification 10 times.

There are three options:

- Keep it as it is
- Do not reset the counts, and instead of asserting that a notification is not sent, assert that it hasn't changes (so still times sent = 10 from the previous sub-test)
- Split the test into 3 tests

In my opinion, one test is sufficient and there is no benefit to splitting it into 3 tests. I also believe it is cleaner / more readable the way that it currently is. I am asserting that a notification is not sent and it's clear why, where as it wouldn't be as clear if I was asserting that it was still sent 10 times.

Let me know your thoughts.

atoff: I personally think what's here is fine. I did a bit of digging to see if there was a way to reset the counter that doesn't look like re-faking and couldn't easily find one.

atoff (approved):

atoff:
@CalumTowers if you are happy can you approve?

CalumTowers:
@atoff Im happy but given some of the prod issues we had with duplicate emails being fired (@AxonC has details) can we coordinate a deployment with some relatively close monitoring for the hours and days after it goes live to ensure this is doing exactly what we want please?

CalumTowers:
@AxonC If emails are behaving on the other changes made, could we look at this one with @BenEsen97 please?

I'm happy to merge it after it has been rebased with main and tested locally as a final QA.

app/Console/Commands/WaitingLists/SendAtcTopTenNotification.php
```diff
@@ -0,0 +1,48 @@
...
+use Illuminate\Support\Collection;
+
+class SendAtcTopTenNotification extends Command
+{
+    /**
+     * The console command name.
+     *
+     * @var string
+     */
+    protected $name = 'waitinglists:sendatctoptennotification';
```
AxonC, line 18:
```suggestion
    protected $name = 'waiting-lists:send-atc-topten-notification';
```

app/Console/Commands/WaitingLists/SendAtcTopTenNotification.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public function handle()
+    {
+        WaitingList::atc()->get()->mapWithKeys(function (WaitingList $waitingList) {
+            return [$waitingList->name => $waitingList
+                ->accountsByEligibility(true)
+                ->take(10)
+                ->filter(function ($account) {
+                    return is_null($account->pivot->within_top_ten_notification_sent_at);
+                }),
```
AxonC, line 38:
```suggestion
                ->filter(fn ($account) => is_null($account->pivot->within_top_ten_notification_sent_at));
                }),
```

app/Notifications/Training/WaitingListAtcTopTen.php
```diff
@@ -0,0 +1,52 @@
...
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toMail($notifiable)
+    {
+        return (new MailMessage)
+            ->from('atc-training@vatsim.uk', 'VATSIM UK - ATC Training')
+            ->subject("Within top 10 on the {$this->list_name} waiting list")
```
AxonC, line 49:
```suggestion
            ->subject("Within top 10 on waiting list: {$this->list_name}")
```

PR #2823, opened 2022-12-13
perf: add database level indexes for networkdata_atc table

AxonC:
This is an experiment to address long-standing performance issues with querying lots of network data at any time.

The three indexes added cover:
- Endorsement condition querying
- Waiting list base 12-hour check (think isUK() scope on the networkdata_atc model)
- Querying for UK network data in any other (before or after the disconnected_at check) usecase

There is a potential this will increase disk usage.

The main motivation for these changes is to potentially remove the need for caching on the waiting list eligibility checking
and thus reducing the overall complexity of this bit of the application.

Relevant Sentry Tracing:
- https://sentry.io/organizations/vatsim-uk/performance/summary/spans/db.sql.query:a24f8f1356d21c38/?project=5734564&query=http.method%3AGET&statsPeriod=24h&transaction=%2Fnova-vendor%2Fwaiting-lists-manager%2Faccounts%2F%7BwaitingList%7D%2Factive%2Findex
- https://sentry.io/organizations/vatsim-uk/performance/summary/spans/db.sql.query:3f9aef01b76f6de3/?project=5734564&query=http.method%3AGET&statsPeriod=24h&transaction=%2Fnova-vendor%2Fwaiting-lists-manager%2Faccounts%2F%7BwaitingList%7D

(Percentile performance influenced by the caching taking over in the relevant time periods)

### DIFF
--- a/database/migrations/2022_12_13_131128_index_networkdata_for_account_disconnected_and_callsign.php
+++ b/database/migrations/2022_12_13_131128_index_networkdata_for_account_disconnected_and_callsign.php
@@ -1,0 +1,47 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('networkdata_atc', function (Blueprint $table) {
+            // index to allow query pattern which retrieves an accounts ATC sessions by their disconnected timestamp
+            // and then callsign.
+            $table->index(['account_id', 'disconnected_at', 'callsign']);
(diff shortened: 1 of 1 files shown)


atoff (approved):
Looks good, would be interesting to see if this allows our checks to be dynamic.
I'm assuming the potentially increased storage would be on the database, not core itself?

AxonC:
> Looks good, would be interesting to see if this allows our checks to be dynamic. I'm assuming the potentially increased storage would be on the database, not core itself?

Indeed! Hopefully this will, although I will say I don't in retrospect agree with using connected_at as the filter in the endorsement condition query. 

Aye, just on the DB.

BenEsen97 (approved):
Nice work @AxonC - if merged this should provide a noticeable improvement to user experience in retrieving real time data (and opens doors to add additional data reporting for members in the future)

PR #2824, opened 2022-12-14
feat: add theory exam pass status to waiting list index

AxonC:
Adds a field to the waiting list index page for ATC lists which indicates whether or not a theory exam
pass has been detected for the user in the list.

The waiting list detects which theory exam to look at via a nullable column in the `training_waiting_list`
table. This will need to be filled out on rollout for each respective level of a waiting list.
This relates to the `exam` field on the `theory_results` table.

The theory_results table in CTS is indexed first on student_id so that additional filtering will be done on
a small number of rows in the logic added to `WaitingListAccount`.

Other changes:
- Added a model for theory exam results which maps to the `theory_results` table in CTS.
- Refactored `Bucket.vue` to have a generic true/false indicator without having to rely on the object structure
of a flag
- Moved the title of a bucket outside the bucket component because I had no idea what I thought when I wrote this 3+ years ago. Some big tidy-up is required at some point in time before I have an effigy assembled in my name and placed on a raging bonfire in disgust. 

I decided against caching this immediately until we have real-world performance measurements on whether the additional complexity is needed. How agile of me!

### DIFF
--- a/app/Models/Training/WaitingList/WaitingListAccount.php
+++ b/app/Models/Training/WaitingList/WaitingListAccount.php
@@ -178,17 +178,19 @@ class WaitingListAccount extends Pivot
         return $this->atcHourCheck() && $this->allFlagsChecker() && $this->current_status->name == 'Active';
     }
 
-    public function getTheoryExamPassedAttribute(): bool
+    public function getTheoryExamPassedAttribute(): ?bool
     {
         if ($this->waitingList->department === WaitingList::PILOT_DEPARTMENT || ! $this->waitingList->cts_theory_exam_level) {
-            return false;
+            return null;
         }
 
-        // the internal CTS member column id not the same as the account id
-        // so search on the indexed column cid.
-        $memberId = Member::where('cid', $this->account_id)->first()->id;
+        $result = TheoryResult::forAccount($this->account_id);
 
-        return TheoryResult::where('student_id', $memberId)
+        if (!$result) {
+            return null;
+        }
+
+        return $result
             ->where('exam', $this->waitingList->cts_theory_exam_level)
             ->where('pass', true)->count() > 0;
     }
--- a/app/Models/Training/WaitingList/WaitingListAccount.php
+++ b/app/Models/Training/WaitingList/WaitingListAccount.php
@@ -2,6 +2,8 @@
 
 namespace App\Models\Training\WaitingList;
 
+use App\Models\Cts\Member;
+use App\Models\Cts\TheoryResult;
 use App\Models\Mship\Account;
 use App\Models\NetworkData\Atc;
 use App\Models\Training\WaitingList;
@@ -174,6 +176,21 @@ class WaitingListAccount extends Pivot
         // are all the flags true
         // and is the atc hour check true
         return $this->atcHourCheck() && $this->allFlagsChecker() && $this->current_status->name == 'Active';
+    }
+
+    public function getTheoryExamPassedAttribute() : bool
+    {
+        if ($this->waitingList->department === WaitingList::PILOT_DEPARTMENT || ! $this->waitingList->cts_theory_exam_level) {
+            return false;
+        }
+
+        // the internal CTS member column id not the same as the account id
+        // so search on the indexed column cid.
+        $memberId = Member::where('cid', $this->account_id)->first()->id;
+
+        return TheoryResult::where('student_id', $memberId)
+            ->where('exam', $this->waitingList->cts_theory_exam_level)
+            ->where('pass', true)->count() > 0;
     }
 
     public function setNotesAttribute($value)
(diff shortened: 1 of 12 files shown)


database/factories/Cts/TheoryResultFactory.php
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * Define the model's default state.
+     *
+     * @return array<string, mixed>
+     */
+    public function definition()
+    {
+        return [
+            'student_id' => factory(Member::class)->create()->id,
+            'exam' => $this->faker->randomElement(['S1', 'S2', 'S3']),
```
BenEsen97, line 22: Is there a particular reason this doesn't include the C1 theory exam? I understand it is purely for randon faking, so just curious.

app/Models/Training/WaitingList/WaitingListAccount.php
```diff
@@ -176,6 +178,21 @@ public function getEligibilityAttribute()
         return $this->atcHourCheck() && $this->allFlagsChecker() && $this->current_status->name == 'Active';
     }
 
+    public function getTheoryExamPassedAttribute(): bool
+    {
+        if ($this->waitingList->department === WaitingList::PILOT_DEPARTMENT || ! $this->waitingList->cts_theory_exam_level) {
+            return false;
```
atoff, line 184: I feel like this function should return `bool?` - feels like we should return `null` if no theory exam level set, and display this on the UI (maybe like a - )? Returning false in this case feels like we could give the user the wrong impression.

Equally, why are we specifically excluding pilot lists from being able to do this?

app/Models/Training/WaitingList/WaitingListAccount.php
```diff
@@ -176,6 +178,21 @@ public function getEligibilityAttribute()
...
+        }
+
+        // the internal CTS member column id not the same as the account id
+        // so search on the indexed column cid.
+        $memberId = Member::where('cid', $this->account_id)->first()->id;
+
+        return TheoryResult::where('student_id', $memberId)
```
atoff, line 191: Not sure how much worth it would be in this situation, but we could do this query through a relation on the Cts/Member model with scopes for pass etc

database/migrations/2022_12_14_103127_add_cts_exam_column.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('training_waiting_list', function (Blueprint $table) {
+            $table->string('cts_theory_exam_level')->nullable();
```
atoff, line 17: Nitpicking here, but `cts_theory_exam_level` wasn't immediately obvious to me what it meant. Maybe `cts_theory_exam_code`?

app/Models/Training/WaitingList/WaitingListAccount.php
AxonC: ATC are the only ones using theory exams in CTS at the moment.

I've got it so that only ATC lists have the column included on the front-end, which is why I thought a true/false was going to be sufficient for now.

app/Models/Training/WaitingList/WaitingListAccount.php
AxonC: Only concern I had there is the query might not use the student_id first which isn't a PK but an indexed column

database/migrations/2022_12_14_103127_add_cts_exam_column.php
AxonC: I chose level because that was the corresponding column name from CTS so wanted to keep consistent

app/Models/Training/WaitingList/WaitingListAccount.php
CalumTowers: I feel like this would be better as a TheoryResult::forAccount(1258635); and abstract the knowledge of it being member_id / cid to the model?

app/Models/Training/WaitingList/WaitingListAccount.php
CalumTowers: I'm indifferent here but my initial read (before seeing Alex's comment) was that null would better represent an absence of any requirement. Passing false, you could argue, is wrong and we should force it to true to pass.

app/Models/Training/WaitingList/WaitingListAccount.php
AxonC: Good shout, have moved it to the model on a static method, only performing the first part of the query. 
https://github.com/VATSIM-UK/core/pull/2824/commits/b08f888d7de73b38df763253a77378706fb73835

app/Models/Training/WaitingList/WaitingListAccount.php
AxonC: Have made some adjustments to return null in certain conditions:

https://github.com/VATSIM-UK/core/pull/2824/commits/b08f888d7de73b38df763253a77378706fb73835

CalumTowers (approved):

PR #2897, opened 2023-01-28
feat: TFP Students Role Sync

atoff:
On deployment, add entry into DiscordRoles between the new permission and discord role 1069020923287588944


CalumTowers:
Looks good but think we also need to sort mentors too?

CalumTowers:
> Looks good but think we also need to sort mentors too?

Ignore me it's already there!

CalumTowers (approved):

atoff:
> Looks good but think we also need to sort mentors too?

had a look, mentors seems to be fine. They are synced using "TFP%" wildcard, and I've checked and everyone in CTS with TFP permissions at the moment has the correct core role so looks good to me!

PR #2968, opened 2023-03-12
feat: separate roles for P1/P2 and TFP mentors

AxonC:


CalumTowers (commented):
For consistency let's remove the Pilot Mentor generic role a few lines above your changes. Once done, LGTM

PR #3040, opened 2023-05-02
feat: Discord Role Rules

atoff:


atoff:
@CalumTowers / @AxonC please take a look through this when you get a min. I have looked again locally and tried the "Member" role syncing that didn't work on prod, and it all seems to work as expected. The test for the `syncToDiscord` method also is pretty explicit in testing this case... Not quite sure what else I can do

PR #3063, opened 2023-05-21
Add callsign to viewable feedback

WillShaw1100:
Hi,

I have made a few changes to the publicly accessible feedback page. This is my first 'play around' with core so I'm still learning, any feedback/comments with improvements for next time are much appreciated.

<hr />

I propose we remove the "Message From" which displays the staff member who sent the feedback (which seems to confuse some members) with a more useful "Callsign" to display the position the user was on when the feedback was received. 

image

All the best!

### DIFF
--- a/app/Models/Mship/Feedback/Feedback.php
+++ b/app/Models/Mship/Feedback/Feedback.php
@@ -110,6 +110,13 @@ class Feedback extends Model
         return $this->hasMany(\App\Models\Mship\Feedback\Answer::class);
     }
 
+    public function position()
+    {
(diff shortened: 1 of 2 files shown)


detheridge02:
I take it the callsign is the callsign of the controller / pilot not the callsign of the person leaving the feedback?

WillShaw1100:
> I take it the callsign is the callsign of the controller / pilot not the callsign of the person leaving the feedback?

Correct. It gets the answer the person leaving the feedback provided from the feedback_answers table.

PR #3064, opened 2023-05-21
Get stand status from UKCP

AndyTWF:
Use UKCP's stand status API to display airfield stand status - including information about assignments, requests and reservations.


app/Libraries/UKCP.php
```diff
@@ -140,4 +144,38 @@ public static function getPathForToken($tokenID, $account)
...
+            $body = json_decode($response->getBody()->getContents(), true);
+
+            return tap(
+                collect($body['stands'])->sortBy('identifier', SORT_NUMERIC)->values()->toArray(),
+                fn (array $stands) => Cache::put(
+                    $this->getStandStatusCacheKey($airfield),
+                    collect($stands)->sortBy('identifier', SORT_NUMERIC)->values()->toArray(),
+                    Carbon::parse($body['refresh_at'])
+                )
+            );
```
atoff, line 169: Looks like a candidate for a bit of `Cache::remember()` https://laravel.com/docs/10.x/cache#retrieve-store

app/Http/Controllers/Site/AirportController.php
```diff
@@ -26,24 +33,13 @@ public function show(Airport $airport)
             });
         })->collapse();
 
-        return $this->viewMake('site.airport.view')->with(['airport' => $airport, 'stations' => $stations, 'stands' => $this->loadStandStatus($airport)]);
-    }
-
-    private function loadStandStatus($airport)
-    {
-        $file_path = resource_path().'/assets/data/stands/'.strtolower($airport->icao).'.csv';
```
atoff, line 34: Can you delete the old stand csv's in `resources/assets/data/stands` too

atoff (changes requested):

app/Libraries/UKCP.php
AndyTWF: Unfortunately `remember` doesn't work in this instance, as we'd need to know the `ttl` up-front, whereas the `ttl` is determined by the response we get :)

app/Http/Controllers/Site/AirportController.php
AndyTWF: Yes! d98d22a9

app/Libraries/UKCP.php
atoff: I did a bit of digging, and `remember` does accept `$ttl` to be a Closure of form `fn ($cachedValue) => int`, so you potentially could refactor so that the refresh_at is passed back, but appreciate that might be a bit OTT!

app/Libraries/UKCP.php
```diff
@@ -140,4 +144,38 @@ public static function getPathForToken($tokenID, $account)
...
+                sprintf('%s/api/stand/status?airfield=%s', config('services.ukcp.url'), $airfield),
+                ['timeout' => 5]
+            );
+            $body = json_decode($response->getBody()->getContents(), true);
+
+            return tap(
+                collect($body['stands'])->sortBy('identifier', SORT_NUMERIC)->values()->toArray(),
+                fn (array $stands) => Cache::put(
+                    $this->getStandStatusCacheKey($airfield),
+                    collect($stands)->sortBy('identifier', SORT_NUMERIC)->values()->toArray(),
```
atoff, line 166: Might be being dumb, but is `->sortBy('identifier', SORT_NUMERIC)->values()->toArray()` happening twice here?

AndyTWF: Nope, that's me being dumb! b4a19491

atoff (approved):

app/Libraries/UKCP.php
```diff
@@ -140,4 +144,38 @@ public static function getPathForToken($tokenID, $account)
     {
         return self::TOKEN_PATH_ROOT.$account->id.'/'.$tokenID.'.json';
     }
+
+    public function getStandStatus(string $airfield): array
```
CalumTowers, line 148: Is it worth being explicit that we are after an ICAO here in the variable name?

app/Libraries/UKCP.php
```diff
@@ -140,4 +144,38 @@ public static function getPathForToken($tokenID, $account)
...
+    {
+        try {
+            $cached = Cache::get($this->getStandStatusCacheKey($airfield));
+            if ($cached) {
+                return $cached;
+            }
```
CalumTowers, line 154: I'm sure there's a way of doing this using Cache::remember?

app/Libraries/UKCP.php
```diff
@@ -140,4 +144,38 @@ public static function getPathForToken($tokenID, $account)
...
+        } catch (ClientException $e) {
+            Log::warning("UKCP Client Error {$e->getMessage()} when getting stand status for {$airfield}");
+
+            return [];
+        }
+    }
+
+    private function getStandStatusCacheKey(string $airfield): string
+    {
+        return sprintf(self::STAND_STATUS_CACHE_FORMAT, $airfield);
```
CalumTowers, line 179: Is there much value extract this onto the class rather than just using a magic string? It feels a bit clunky to read.

app/Libraries/UKCP.php
AndyTWF: Yes! fb85e2ea

app/Libraries/UKCP.php
AndyTWF: Yes! 62436dce

app/Libraries/UKCP.php
AndyTWF: Trying it here! 4ec12dce

CalumTowers (approved):
Nice work! 🙌

PR #3091, opened 2023-06-04
fix: handle -1 pilotrating bitmask and pilot rating revokation

AxonC:
This fixes the reported trello bug where P0 members were being granted pilot ratings upto P5. This was due to the bitmask logic when parsing the ratings squaring a negative number (-1) and then shifting all the way upto P5.

This also supports pilot ratings for Flight Instructor & Flight Examiner via UpdateMember and allows the new behaviour when these ratings are revoked in favour of previous permeanant pilot ratings.


detheridge02:
All looks good to me

PR #3111, opened 2023-06-19
feat: add feature toggles to waiting lists

AxonC:


app/Models/Training/WaitingList.php
```diff
@@ -175,6 +179,28 @@ public function getFormattedDepartmentAttribute()
         };
     }
 
+    public function getShouldCheckAtcHoursAttribute(): bool
+    {
+        // default to true
+        return $this->isAtcList() && ($this->feature_toggles['check_atc_hours'] ?? true);
```
CalumTowers, line 185: Why is this not just looking at the toggle? Would be much cleaner and obvious to the user as to what is going on. Also would remove the need for the comment.

app/Models/Training/WaitingList.php
```diff
@@ -175,6 +179,28 @@ public function getFormattedDepartmentAttribute()
...
+    }
+
+    public function getShouldCheckCtsTheoryExamAttribute(): bool
+    {
+        // default to true for atc lists.
+        return $this->isAtcList() && ($this->feature_toggles['check_cts_theory_exam'] ?? true);
```
CalumTowers, line 191: Why is this not just looking at the toggle? Would be much cleaner and obvious to the user as to what is going on. Also would remove the need for the comment.

app/Models/Training/WaitingList.php
```diff
@@ -175,6 +179,28 @@ public function getFormattedDepartmentAttribute()
...
+    }
+
+    public function getFeatureTogglesFormattedAttribute(): object
+    {
+        // convert feature toggles to an object with defaults included to
+        // ensure keys are always present.
```
CalumTowers, line 197: I'm not sure this comment is helpful, the code is fairly verbose

app/Models/Training/WaitingList/WaitingListAccount.php
```diff
@@ -140,6 +140,10 @@ public function atcHourCheck()
             return true;
         }
 
+        if ($this->waitingList->should_check_atc_hours == false) {
```
CalumTowers, line 143:
```suggestion
        if (! $this->waitingList->should_check_atc_hours) {
```

CalumTowers (changes requested):

app/Models/Training/WaitingList.php
AxonC: Historically we have only shown cts theory exams & atc hour checks on ATC lists so this is just maintaining backwards compatibility.

app/Models/Training/WaitingList.php
AxonC: changed

app/Models/Training/WaitingList.php
AxonC: changed

PR #3129, opened 2023-06-28
feat: Filament Account Management

atoff:
Closes Linear WEB-126


atoff:
@AxonC this is probably ready for review when you feel up to it 😂 Majority of the files changed are refactors

AxonC:
> @AxonC this is probably ready for review when you feel up to it 😂 Majority of the files changed are refactors

All LGTM - I would like to still have the ability to seed the old local permissions locally. On that point, have we checked the new and the old can co-exist?

AxonC (approved):

atoff:
> > @AxonC this is probably ready for review when you feel up to it 😂 Majority of the files changed are refactors
> 
> All LGTM - I would like to still have the ability to seed the old local permissions locally. On that point, have we checked the new and the old can co-exist?

I've kept the old permissions in the migrations - these seemed to be duplicated in the seeder and would be in the database upon migration. Once we are done with the old permissions, we should remove them from the migrations (and prod DB manually).

Shouldn't be any conflicts, and even if there were the seeder is setup to only create if the permission doesn't exist, so should be fine!

AxonC (approved):

BenEsen97 (approved):

atoff:
@CalumTowers please could the following be added to the Forge deployment, after `php artisan migrate`:
```
php artisan db:seed --class=RolesAndPermissionsSeeder
```

PR #3137, opened 2023-07-05
refactor Dockerfile for devcontainer

detheridge02:
Updated to use base of webdevops/php-nginx-dev:8.2 to support PHP 8.2.x.
This is required following changes to docker-compose.yml


detheridge02:
Investigating further the Composer / PHP 8.1 issues

PR #3138, opened 2023-07-06
feat: waiting list eligibility is database & event driven

AxonC:
This PR fundamentally changes the way in which we determine the eligibility of accounts within a waiting list.

Previously, we used a computed 'accessor' attribute contained in the pivot model `WaitingListAccount`. This resulted in some complexity with caching and performance as it wasn't an actual data attribute. This results in a very long process with caching the results and a state for members where the result of ATC hours does not get instantly reflected. In addition, with a view to moving away from Nova, it would be advantageous to be able to query just on the pivot attribute without always having to do some computation.

This PR, in a nutshell, does the following:
- Removes the `eligibility` 'accessor' property on `WaitingListAccount` in favour of new database column `eligible`.
- Populates the `eligible` property every 24 hours via Artisan command which creates jobs on a queue to check an accounts eligibility. In addition, it populates two further columns on how that `eligible` property was determined.
- Adds event listener to generically take a new `AccountCentricEvent` which is implemented by a number of WaitingList events, including the events for when an account is added to a waiting list, a manual flag is toggled, a flag is added at large to a waiting list & more importantly when an ATC session has ended.


app/Console/Commands/Training/CheckWaitingListEligibilityCommand.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $activeWaitingLists = WaitingList::all();
+
+        if ($this->argument('account')) {
+            $activeWaitingLists = $activeWaitingLists->filter(function ($waitingList) {
+                return $waitingList->accounts->contains($this->argument('account'));
+            });
+
+            if ($activeWaitingLists->isEmpty()) {
+                $this->error('Account not found on any waiting lists');
```
CalumTowers, line 40: This error is probably not right. Something like “no active waiting lists found”?

app/Events/Training/AccountManualFlagChanged.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Get the channels the event should broadcast on.
+     *
+     * @return \Illuminate\Broadcasting\Channel|array
+     */
+    public function broadcastOn()
+    {
+        return new PrivateChannel('channel-name');
+    }
```
CalumTowers, line 36: Can this be removed?

app/Events/Training/FlagAddedToWaitingList.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Get the channels the event should broadcast on.
+     *
+     * @return \Illuminate\Broadcasting\Channel|array
+     */
+    public function broadcastOn()
+    {
+        return new PrivateChannel('channel-name');
+    }
```
CalumTowers, line 32: Can this be removed?

CalumTowers (commented):
Looks good!

app/Services/Training/CheckWaitingListEligibility.php
```diff
@@ -0,0 +1,82 @@
...
+            ->sum('minutes_online');
+
+        $this->baseControllingHoursCheck = $recentAtcMinutes >= 720;
+
+        return $this->baseControllingHoursCheck;
+    }
+
+    public function checkWaitingListFlags(WaitingList $waitingList)
+    {
+        $waitingListAccount = $waitingList->accounts()->where('account_id', $this->account->id)->first()->pivot;
```
atoff, line 46: Shouldn't `account_id` be `id` here because it's querying `App\Models\Mship\Account` models not `WaitingListAccount` pivot models? I guess it is using the pivot model attribute in this case, but might be best to stick to `id` for convention?

app/Services/Training/CheckWaitingListEligibility.php
```diff
@@ -0,0 +1,82 @@
...
+
+        if ($waitingList->flags->count() == 0) {
```
atoff, line 48:
```suggestion
        if ($waitingList->flags()->doesntExist()) {
```

app/Services/Training/CheckWaitingListEligibility.php
```diff
@@ -0,0 +1,82 @@
...
+        $recentAtcMinutes = Atc::where('account_id', $this->account->id)
+            ->where('disconnected_at', '>=', Carbon::parse('3 months ago'))->isUk()
+            ->sum('minutes_online');
+
+        $this->baseControllingHoursCheck = $recentAtcMinutes >= 720;
+
+        return $this->baseControllingHoursCheck;
+    }
+
+    public function checkWaitingListFlags(WaitingList $waitingList)
```
atoff, line 44: Could we add some docblocs to these functions just to briefly explain what they do, and the return data structure

nova-components/WaitingListsManager/src/Http/WaitingListAccountResource.php
```diff
@@ -16,9 +16,9 @@ public function toArray($request)
             'id' => $this->id,
             'pivot_id' => $this->pivot->id,
             'name' => $this->name,
-            'atcHourCheck' => $this->pivot->atcHourCheck,
+            'atcHourCheck' => $this->pivot->eligibility_summary['base_controlling_hours'] ?? false,
```
atoff, line 19: Maybe use a graceful array accessor here to prevent any exceptions if the index doesn't exist?
`Arr::get($this->pivot->eligibility_summary, 'base_controlling_hours', false)`

resources/views/mship/waiting-lists/view.blade.php
```diff
@@ -89,7 +89,7 @@
 </div>
 @if($list->isATCList() || count($automaticFlags))
 <div class="alert alert-warning">
-    <strong>Important: </strong> Automated eligibility flags are only calculated every 24 hours! If you have just completed
+    <strong>Important: </strong> Automated eligibility flags are only calculated every 24 hours, or after the end of an ATC session! If you have just completed
     a network session, the flags shown above may not be accurate.
```
atoff, line 93: Might need to remove the last part of this paragraph if we imply it does update after ATC session? Or is it being kept to say it might not be instantaneous

atoff (changes requested):

app/Console/Commands/Training/CheckWaitingListEligibilityCommand.php
AxonC: This only executes when an account is specified, so I think this is okay?

app/Events/NetworkData/AtcSessionEnded.php
```diff
@@ -21,4 +23,9 @@ public function __construct(Atc $atc)
     {
         $this->atcSession = $atc;
     }
+
+    public function getAccount(): Account
+    {
+        return $this->atcSession->load('account')->account;
```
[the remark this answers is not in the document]
AxonC: Avoids N + 1 query problem

app/Models/Training/WaitingList.php
```diff
@@ -79,9 +83,7 @@ public function accountsByEligibility($eligible = true)
         return $this->accounts()
             ->orderByPivot('created_at')
             ->get()
-            ->filter(function ($model) use ($eligible) {
-                return $model->pivot->eligibility == $eligible;
-            })->values();
+            ->filter(fn ($model) => $model->pivot->eligible == $eligible)->values();
```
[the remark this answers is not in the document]
AxonC: I have that coming in the next PR for two relationships to create Eligible and Ineligible relation managers

app/Services/Training/CheckWaitingListEligibility.php
AxonC: Sadly it's ambiguous so account_id is needed

app/Console/Commands/Training/CheckWaitingListEligibilityCommand.php
```diff
@@ -0,0 +1,66 @@
...
+     * @return int
+     */
+    public function handle()
+    {
+        $activeWaitingLists = WaitingList::all();
+
+        if ($this->argument('account')) {
+            $activeWaitingLists = $activeWaitingLists->filter(function ($waitingList) {
+                return $waitingList->accounts->contains($this->argument('account'));
+            });
```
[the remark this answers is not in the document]
AxonC: have moved stuff around so it's no longer always a collection full of models so nae can do

app/Console/Commands/Training/CheckWaitingListEligibilityCommand.php
```diff
@@ -0,0 +1,66 @@
...
+            $activeWaitingLists = $activeWaitingLists->filter(function ($waitingList) {
+                return $waitingList->id === $this->option('waiting-list');
+            });
+
+            if ($activeWaitingLists->isEmpty()) {
+                $this->error('Waiting list not found');
+
+                return Command::FAILURE;
+            }
+        }
```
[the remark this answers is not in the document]
AxonC: good idea

app/Services/Training/CheckWaitingListEligibility.php
AxonC: done!

app/Services/Training/CheckWaitingListEligibility.php
AxonC: done!

nova-components/WaitingListsManager/src/Http/WaitingListAccountResource.php
AxonC: nice one!

app/Jobs/Training/CheckAccountWaitingListEligibility.php
```diff
@@ -0,0 +1,42 @@
...
+use App\Models\Mship\Account;
+use App\Services\Training\CheckWaitingListEligibility;
+use App\Services\Training\WriteWaitingListEligibility;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+
+class CheckAccountWaitingListEligibility implements ShouldQueue
```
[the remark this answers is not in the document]
AxonC: done!

resources/views/mship/waiting-lists/view.blade.php
AxonC: amended slightly

PR #3150, opened 2023-07-09
feat: add feedback to filament admin panel

AxonC:
This PR:

- Adds Feedback resource to filament admin panel
- Adds filament package for copying fields
- Changes policies to look at both old and new permission styles to maintain backward compatibility whilst Nova and Filament co-exist


CalumTowers (approved):
Looks great to me.

PR #3281, opened 2023-10-25
Make contributing slightly easier

maxbrokman:
### Changes

- Adds readme
- Adds `composer build` command
- Moves `phpunit.xml` to `phpunit.xml.dist`

For details see below or individual commit messages

### Target

Targets `filament-v3-upgrade` for now because building the project on `main` is not possible without a Laravel Nova account, which I don't have!

PR is left in Draft state for this reason.

### Motivation

I was interested by conversation about #3276 and noticed that that PR does not include any tests.

When setting out to fix this I noticed there was no readme on this repo so decided to add one. In doing so I have also made a couple of changes with the aim of making getting the project up and running more straightforward. 

### Detailed Changes

**Note**: I am running composer & yarn from the provided dockerimage. I did not directly use the devcontainer functionality.

#### Readme

Adds a basic readme with the same information as that found on [Nexus](https://github.com/VATSIM-UK/nexus/blob/main/README.md). The objective of the readme provided is to get the developer to the point of being able to run the tests. This seems to have been successful, with two tests skipped.

#### Move to `phpunit.xml.dist`

IMO best practice is to distribute a `.dist` file to allow individuals to override any phpunit settings without risk of accidentally committing this to vcs.

As part of this change I have also added editorconfig entries for .xml and .xml.dist which sets a 2 space identation, in order to match existing files.

#### Adds `composer build` command

This simply migrates the databases and publishes the Horizon & Telescope vendor f/e code. Seemed convenient and is done by the tests in CI so also seemed important.

### Note on Yarn, Telescope, Horizon

I noticed that in order to build the frontend I had to let yarn (using the version distributed in the dockerfile) generate a new lock file. I have not committed this but it seemed relevant given the deploy workflow does not make use of a frozen lockfile. 

Similarly publishing f/e code from Telescope & Horizon changes files under version control, though these are run before tests in CI. I'm not sure if these files should be in VCS or not but my gut feeling is that they should not be.

I am aware that using this `composer build` command may lead to accidental changes being committed and you may wish to remove / amend it as you see fit.


detheridge02:
Great work. 
Once the filament PR has been approved I'll run through this to install my environment and let you know if any changes are required

maxbrokman:
Great thanks

PR #3314, opened 2023-11-01
feat: Notes on Account Resource

atoff:


app/Filament/Resources/AccountResource/RelationManagers/NotesRelationManager.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    protected static function getSubject($livewire): Account
+    {
+        return $livewire->ownerRecord;
+    }
+
+    protected static function getAttachment($livewire): ?Model
+    {
+        return null;
```
AxonC, line 77: Intended?

AxonC (approved):

app/Filament/Resources/AccountResource/RelationManagers/NotesRelationManager.php
atoff: General notes don't need an attachment - I've implemented that method so that the Ban version of the relationmanager can redefine it to return the ban it's attached to!

PR #3399, opened 2024-01-02
Add TC Mids to top down order

PLM1995:
Per https://community.vatsim.uk/topic/39092-2023-12-28-aerodromes-egllssgwnxbbtktc-changes-to-coordination-procedures-due-to-tc-midlands-implementation/

Added EGTC and EGTK to top-down order list
Added TC Mids to who to call


KHardern (approved):
Correct from an ops perspective.

PR #3465, opened 2024-02-13
feat: Prefill usercid in feedback forms

maxbrokman:
A little poc for prefilling the usercid in e.g the atc feedback form.

Saw some discussion on discord about trying to search for a controller name in order to submit feedback, I know I've been asked by pilots what my CID is so they can submit feedback about me (good feedback I hope!)

This fills the `Question` with slug `usercid` if the `usercid` is present in the query params.

Because the link to the new atc feedback form is quite long and getting a query param through a url shortener (`vats.im/atcfb`) is challenging I've also added a short url to the atc feedback form (`vatsim.uk/atcfb`)

My intention would be to use it in euroscope profiles as `Submit feedback at vatsim.uk/atcfb?usercid=12345678`).

Keen to know your thoughts, further notes below.

-------------

I believe the url shortener is short.io? It would be possible to generate unique short urls for each controller, but this would be way more complex so I've just done this for now.

-------------

I tried using a redirect route but this had two problems:
1. I'm not sure if this would pass query params along, but couldn't test it as
2. Defining a top level redirect to a named route seems to be impossible when the route is defined in a group closure, guessing laravel evaluates redirects before groups.

### DIFF
--- a/app/Http/Controllers/Mship/Feedback.php
+++ b/app/Http/Controllers/Mship/Feedback.php
@@ -199,4 +199,9 @@ class Feedback extends \App\Http\Controllers\BaseController
 
         return response()->json($this->returnList);
     }
+
+    public function redirectNewAtc(Request $request)
+    {
+        return redirect()->route('mship.feedback.new.form', array_merge(['atc'], $request->query->all()));
+    }
 }
--- a/app/Http/Controllers/Mship/Feedback.php
+++ b/app/Http/Controllers/Mship/Feedback.php
@@ -44,8 +44,9 @@ class Feedback extends \App\Http\Controllers\BaseController
     public function getFeedback(Form $form, Request $request)
     {
         // Get a list of "pre-fillable" questions from the request, based on question slug
-        // To allow direct links from e.g euroscope profiles via vats.im/atcfb?usercid=12345
-        $preFillable = $request->only(['usercid']);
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Mship/Feedback.php
```diff
@@ -40,8 +41,13 @@ public function postFeedbackFormSelect(Request $request)
         return Redirect::route('mship.feedback.new.form', [$request->input('feedback_type')]);
     }
 
-    public function getFeedback(Form $form)
+    public function getFeedback(Form $form, Request $request)
     {
+        // Get a list of "pre-fillable" questions from the request, based on question slug
+        // To allow direct links from e.g euroscope profiles via vats.im/atcfb?usercid=12345
+        $preFillable = $request->only(['usercid']);
```
AxonC, line 48: Minor point: I'd ask to keep the nomenclature for 'user' inline with the Core naming for an account.

```suggestion
        $preFillable = $request->only(['accountcid']);
```

Or just use `cid` as the query parameter.

maxbrokman: Have swapped to `cid` as it's shorter

CalumTowers (approved):
Thanks for the contribution @maxbrokman! I've made a few tweaks and will merge this in 👍

maxbrokman:
@hazzas-99 it's been released as https://vatsim.uk/atcfb?cid=12345 in the end which should work

PR #3565, opened 2024-04-02
feat: sync new S1 exam passes to roster for home members

AxonC:


app/Console/Commands/Roster/CheckForNewS1ExamPasses.php
```diff
@@ -0,0 +1,62 @@
...
+                continue;
+            }
+
+            if (! $coreAccount->hasState(State::findByCode('DIVISION'))) {
+                $this->error("Account {$coreAccount->id} does not have the DIVISION state.");
+
+                continue;
+            }
+
+            $isAlreadyOnRoster = Roster::where('account_id', $coreAccount->id)->exists();
```
CalumTowers, line 51: Use upsert here to simplify.

AxonC: good shout - done!

PR #3579, opened 2024-04-05
Update Allowed Display Names to Include CID

CalumTowers:


AxonC:
Can we update the default creation of the bookmark to include the name & CID?

PR #3581, opened 2024-04-07
TECH-245 Gatwick Endorsement Requirements Page

maxbrokman:
Re-Implement Gatwick Endorsement requirements page for the new requirements, and update some language as necessary.

Endorsement page checks for

- 50 hours as DEL or GND in the UK based on Network Atc role data
- Home division membership
- Active on Roster
- Holding an S1

All working based on tests / qa, would appreciate a review of the language in the requirement descriptions.

### DIFF
--- a/app/Http/Controllers/BaseController.php
+++ b/app/Http/Controllers/BaseController.php
@@ -21,6 +21,7 @@ class BaseController extends \Illuminate\Routing\Controller
     }
     use DispatchesJobs, RedirectsUsers, ValidatesRequests;
 
+    /** @var Account */
     protected $account;
 
     protected $pageTitle;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/BaseController.php
```diff
@@ -21,6 +21,7 @@ class BaseController extends \Illuminate\Routing\Controller
     }
     use DispatchesJobs, RedirectsUsers, ValidatesRequests;
 
+    /** @var Account */
     protected $account;
```
AxonC, line 25: Nitpick: 
```suggestion
    protected Account $account;
```

AxonC (approved):

app/Http/Controllers/BaseController.php
maxbrokman: Crashes elsewhere due to access before initialisation annoyingly, will dig in to that in a separate or potentially

PR #3582, opened 2024-04-08
feat: notify users upon being granted a permanent endorsement aka. PositionGroup

AxonC:
This establishes a pattern for sending events from Endorsement events and adds a notification
to inform a user they have been endorsed on a PositionGroup aka. Tier 1 / 2 endorsement.

image

### DIFF
--- a/app/Events/Mship/Endorsement/TierEndorsementAdded.php
+++ b/app/Events/Mship/Endorsement/TierEndorsementAdded.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace App\Events\Mship\Endorsement;
+
+use App\Models\Mship\Account;
+use App\Models\Mship\Account\Endorsement;
+use Illuminate\Broadcasting\InteractsWithSockets;
+use Illuminate\Broadcasting\PrivateChannel;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Queue\SerializesModels;
+
+class TierEndorsementAdded
+{
+    use Dispatchable, InteractsWithSockets, SerializesModels;
+
+    /**
+     * Create a new event instance.
+     */
+    public function __construct(private Endorsement $endorsement, private Account $account)
+    {
+    }
+
+    /**
+     * Get the channels the event should broadcast on.
+     *
+     * @return array<int, \Illuminate\Broadcasting\Channel>
+     */
+    public function broadcastOn(): array
+    {
+        return [
+            new PrivateChannel('channel-name'),
+        ];
+    }
+
+    public function getAccount()
+    {
+        return $this->account;
+    }
+
+    public function getEndorsement()
+    {
+        return $this->endorsement;
+    }
+}
--- a/app/Events/Mship/Endorsement/TierEndorsementAdded.php
+++ b/app/Events/Mship/Endorsement/TierEndorsementAdded.php
@@ -5,7 +5,6 @@ namespace App\Events\Mship\Endorsement;
 use App\Models\Mship\Account;
 use App\Models\Mship\Account\Endorsement;
 use Illuminate\Broadcasting\InteractsWithSockets;
-use Illuminate\Broadcasting\PrivateChannel;
 use Illuminate\Foundation\Events\Dispatchable;
 use Illuminate\Queue\SerializesModels;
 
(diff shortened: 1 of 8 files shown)


app/Events/Mship/Endorsement/TierEndorsementAdded.php
```diff
@@ -0,0 +1,44 @@
...
+     * Get the channels the event should broadcast on.
+     *
+     * @return array<int, \Illuminate\Broadcasting\Channel>
+     */
+    public function broadcastOn(): array
+    {
+        return [
+            new PrivateChannel('channel-name'),
+        ];
+    }
```
CalumTowers, line 33: Not required

app/Providers/EventServiceProvider.php
```diff
@@ -110,6 +110,9 @@ class EventServiceProvider extends ServiceProvider
         DiscordUnlinked::class => [
             RemoveDiscordUser::class,
         ],
+        \App\Events\Mship\Endorsement\TierEndorsementAdded::class => [
+            \App\Listeners\Mship\Endorsement\NotifyOfTierEndorsement::class,
+        ],
```
CalumTowers, line 115: Import rather than fully qualified

resources/views/emails/mship/endorsement/tier_endorsement_granted.blade.php
```diff
@@ -0,0 +1,16 @@
...
+@section('body')
+<p>This email is formal notification that you have been granted a 'Tier' endorsement: <strong>{{ $endorsement_name }}</strong>.</p>
+
+<p>This endorsement entitles you additional permissions to control the following positions:</p>
+<ul>
+	@foreach($positions as $position)
+		<li>{{ $position }}</li>
+	@endforeach
+</ul>
+<p>The listed positions are inn addition to the positions you are already entitled to control via your controller rating and other Tier or Solo endorsements.
```
CalumTowers, line 12: "inn" should be in "in"

CalumTowers (changes requested):

app/Providers/EventServiceProvider.php
AxonC: General convention in that file seems to be full path if you look at 95% of the mappings. Happy to go either way but the file would be 95% imports

CalumTowers: One of these days I'll just import everything across the project! Fully qualified looks messy

CalumTowers (approved):

app/Events/Mship/Endorsement/TierEndorsementAdded.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Queue\SerializesModels;
+
+class TierEndorsementAdded
+{
+    use Dispatchable, InteractsWithSockets, SerializesModels;
+
+    /**
+     * Create a new event instance.
+     */
+    public function __construct(private Endorsement $endorsement, private Account $account)
```
CalumTowers, line 18: Why are we passing both the endorsement and the account? Why not just accept the endorsement access the account from the endorsement?

Could they also be public so that they can be accessed without the getters?

app/Observers/EndorsementObserver.php
```diff
@@ -0,0 +1,56 @@
...
+     * Handle the Endorsement "created" event.
+     */
+    public function created(Endorsement $endorsement): void
+    {
+        $endorsement->load('account');
+
+        if ($endorsement->endorsable_type == PositionGroup::class) {
+            event(new TierEndorsementAdded($endorsement, $endorsement->account));
+
+            return;
```
CalumTowers, line 21: Is this return needed? Wouldn't it just return void anyway?

app/Observers/EndorsementObserver.php
```diff
@@ -0,0 +1,56 @@
...
+class EndorsementObserver
+{
+    /**
+     * Handle the Endorsement "created" event.
+     */
+    public function created(Endorsement $endorsement): void
+    {
+        $endorsement->load('account');
+
+        if ($endorsement->endorsable_type == PositionGroup::class) {
```
CalumTowers, line 18: Might look nicer as a match statement to make future expansion easier

CalumTowers (commented):
Some other thoughts as I take another look

app/Events/Mship/Endorsement/TierEndorsementAdded.php
AxonC: The account is eager loaded in the event constructor on the observer. The separation was just a bit more explicit to stop an N + 1.

app/Observers/EndorsementObserver.php
AxonC: Agreed - another change to come soon for solo endorsements where pattern matching is going to be super helpful.

PR #3634, opened 2024-05-06
feat: send notification to approvers when endorsement request created

AxonC:


app/Notifications/Mship/Endorsement/EndorsementRequestCreated.php
```diff
@@ -0,0 +1,48 @@
...
+     * Get the mail representation of the notification.
+     */
+    public function toMail(object $notifiable): MailMessage
+    {
+        return (new MailMessage)
+            ->from(config('mail.from.address'), 'VATSIM UK - Training Department')
+            ->subject('Endorsement Request Created')
+            ->view('emails.mship.endorsement.endorsement_request_created', [
+                'recipient' => $notifiable,
+                'requestIndexUrl' => url('/admin/endorsement-requests'),
```
CalumTowers, line 42: Can we use the route helper to ensure that if the URI changes this updates too?

CalumTowers (approved):

app/Notifications/Mship/Endorsement/EndorsementRequestCreated.php
AxonC: Sadly filament is only seen as a single 'route' (`filament.update`), so a bit of hard coding is needed here.

atoff: Only just saw this, but fwiw should be able to do something like
```php
                'requestIndexUrl' => EndorsementRequestResource::getUrl('index'),
```

PR #3671, opened 2024-06-12
TECH-247 Waiting list performance

maxbrokman:
Number of queries loading a 200 person waiting list reduced from ~200 to ~10, will hopefully help with some of the crashes on the waiting lists. More details below.

**due to lack of familiarity could someone do some QA on this with a more real dataset**. For testing I generated 200 fake accounts and worked with that but I wouldn't say I'm confident especially in adding / removing people from waiting lists.

I would like to come back and rework some of the relationships / clean up but have left things tagged deprecated for now as this is already a pretty large change.

---

fix: Reduce number of queries being run by filament when viewing waiting lists

Moving WaitingList <> Account from a ManyMany to two BelongsTo relationships means we can avoid N+1 problems in filament.

This approach dramatically reduces the number of database queries being run when viewing a waiting list. Query count reduced by an order of magnitude, should help with performance / 500s.

Also for eager loading purposes there is now a relationship between an account and roster status, but this is only being used in waiting list filament currently


app/Providers/AppServiceProvider.php
```diff
@@ -62,6 +65,10 @@ public function boot()
         Cookies::essentials()
             ->session()
             ->csrf();
+
+        DB::listen(function (QueryExecuted $query) {
+            Log::debug($query->sql, $query->bindings);
+        });
```
AxonC, line 71: Is this intentional in the final changeset?

app/Models/Training/WaitingList.php
```diff
@@ -61,6 +105,10 @@ public function staff()
 
     /**
      * Many WaitingLists can have many Accounts (pivot).
+     *
+     * @fixme remove when no longer used in filament stuff, use `waitingListAccounts` instead.
```
AxonC, line 109: Does this still apply?

app/Providers/AppServiceProvider.php
maxbrokman: noooo 😄

app/Models/Training/WaitingList.php
maxbrokman: yes i think so, i believe this is used somewhere within adding / removing someone in filament, i'll have another look

app/Models/Mship/Account.php
```diff
@@ -365,6 +420,16 @@ public function addNote($noteType, $noteContent, $writer = null, $attachment = n
...
     }
 
+    public function roster(): HasOne
+    {
+        return $this->hasOne(Roster::class, 'account_id', 'id');
+    }
+
+    public function onRoster(): bool
+    {
+        return $this->roster?->account_id == $this->id;
```
AxonC, line 430: Can you just query the relationships existence instead of a null attribute check?

app/Models/Training/WaitingList.php
```diff
@@ -101,6 +162,16 @@ public function accountPosition(Account $account)
         return ($key !== false) ? $key + 1 : null;
     }
 
+    /**
+     * Alternative to accountPosition for use with filament, beware N+1 issues
+     */
+    public function positionOf(WaitingListAccount $waitingListAccount): ?int
```
AxonC, line 168: Nitpick: Can't find a test for this method directly

AxonC (approved):
LGTM - couple of minor things

app/Models/Mship/Account.php
maxbrokman: so I originally had this as `$this->roster->exists()` but that actually causes a query to be issued. Could do a `!== null`

AxonC: Fine to leave as is then!

PR #3689, opened 2024-06-22
remove /admin from UKCP link

kristiankunc:
Removes `/admin` from the UKCP url.
Same issue as in https://github.com/VATSIM-UK/core/pull/3655 I've just missed this second url


CalumTowers:
Thanks @kristiankunc

PR #3698, opened 2024-07-02
TECH-278 Waiting List Position 

maxbrokman:
Restore viewing waiting list position for members.

Planning to come back and do some big cleanups on waiting list stuff (mix of double relation / pivot still remains)

Removes view waiting list details page as requirements for atc are much simpler now (stay on roster)`


app/Events/Training/AccountAddedToWaitingList.php
```diff
@@ -16,7 +16,9 @@ class AccountAddedToWaitingList implements AccountCentricEvent
     /**
      * Create a new event instance.
      */
-    public function __construct(public Account $account, public WaitingList $waitingList, public Account $staffAccount) {}
+    public function __construct(public Account $account, public WaitingList $waitingList, public Account $staffAccount)
+    {
```
maxbrokman, line 20: linter attack

CalumTowers (approved):

PR #3699, opened 2024-07-03
Housekeeping

CalumTowers:
I've been pondering about starting a bit of housekeeping on Core for a while.

There are a lot of things that are left over from times gone by and some of them can easily be removed with minimal fuss - as they are no longer used.

This PR starts that by:-

- Removing `HerokuPostDeploy`, from when we used to deploy PRs to Heroku
- Moving some of our development-only commands to a dedicated namespace
- Removing a command to generate eloquent PHP docs
- Removing a command to test emails
- Removing the syncing of TG forum groups which runs inconsistently in production and is no longer required
- Removing a feedback summary generator that has not been used for many years
- Moving a number of VATSIM.net services from their own config files to the services config file
- Removing a command to reprocess bids, as it was only used once or twice when we deployed some fixes many years ago
- Remove an unused boolean indicator blade component
- Remove the unit test stub


stubs/test.unit.stub
```diff
@@ -1,16 +0,0 @@
-<?php
-
-namespace {{ namespace }};
-
-use Tests\TestCase;
-
-class {{ class }} extends TestCase
```
AxonC, line 7: I added this originally because the default  test case imports from the PHPUnit test case not our custom one.

AxonC: I should've put a docstring to that effect on reflection

AxonC (approved):
One comment otherwise LGTM!

stubs/test.unit.stub
CalumTowers: 7de5c825babd642ab9cb07758a299dfab2b71bb5

PR #3852, opened 2024-11-04
refactor: Move WaitingListAccount to entity

maxbrokman:
Quite a large change set I'm afraid!

Moves WaitingListAccount from a pivot table to a full fat entity. This enables further cleanup work and hopefully makes introducing further waiting list features a bit more straightforward. Having a pivot table meant access to waiting list position information was a bit ugly in places and in the past has lead to some big N+1 problems.

The only major difference is the way waiting lists are accessed from accounts and vice-versa is different now, involving a method call instead of using laravel relationships.


CalumTowers:
@AxonC You're probably best placed to review this

AxonC (approved):

PR #3947, opened 2025-01-07
Improved Stats

CalumTowers:
- Adds stats to quarterly stats for ATC related information
- Adds logging of roster updates and key stats on each run, which are in turn displayed in admin

Resolves TECH-310
Resolves TECH-261

### DIFF
--- a/app/Filament/Resources/RosterUpdateResource.php
+++ b/app/Filament/Resources/RosterUpdateResource.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace App\Filament\Resources;
+
+use App\Filament\Resources\RosterUpdates\Pages\ListRosterUpdates;
+use App\Filament\Resources\RosterUpdates\Pages\ViewRosterUpdate;
+use App\Models\RosterUpdate;
+use App\Models\Training\WaitingList;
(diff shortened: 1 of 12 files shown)


app/Filament/Resources/RosterUpdateResource.php
```diff
@@ -0,0 +1,82 @@
...
+use Filament\Forms\Components\DatePicker;
+use Filament\Forms\Components\Section;
+use Filament\Forms\Components\TextInput;
+use Filament\Forms\Form;
+use Filament\Resources\Resource;
+use Filament\Tables\Actions\ViewAction;
+use Filament\Tables\Columns\TextColumn;
+use Filament\Tables\Table;
+
+class RosterUpdateResource extends Resource
```
AxonC, line 18: Do you need any permissions behind this?

AxonC: i.e. a policy etc.. Could use the existing roster.manage permission?

CalumTowers: Good point - added in 91a1016

AxonC (approved):

PR #3948, opened 2025-01-09
TECH-232: Add support for VATSIM.net Webhooks

CalumTowers:
Docs: https://vatsim.dev/services/webhooks.

_If testing locally on `sync`, be sure to remove `->afterResponse()`._


app/Http/Controllers/External/VatsimNet/ProcessVatsimNetWebhook.php
```diff
@@ -0,0 +1,28 @@
...
+        if (request()->header('Authorization') !== config('services.vatsim-net.webhook.key')) {
+            return response()->json([
+                'status' => 'forbidden',
+            ], 403);
+        }
+
+        foreach (request()->json('actions') as $action) {
+            $class = config("services.vatsim-net.webhook.jobs.{$action['action']}");
+            if ($class && class_exists($class)) {
+                dispatch(new $class(request()->json('resource'), $action))->afterResponse();
```
maxbrokman, line 20: for ease of running sync locally here can we just drop `->afterResponse()` entirely. Dispatching queue jobs should be fast anyway and I don't think the system sending the webhooks is gonna mind too much

```suggestion
                dispatch(new $class(request()->json('resource'), $action));
```

app/Http/Controllers/External/VatsimNet/ProcessVatsimNetWebhook.php
```diff
@@ -0,0 +1,28 @@
...
+    public function __invoke()
+    {
+        if (request()->header('Authorization') !== config('services.vatsim-net.webhook.key')) {
+            return response()->json([
+                'status' => 'forbidden',
+            ], 403);
+        }
+
+        foreach (request()->json('actions') as $action) {
+            $class = config("services.vatsim-net.webhook.jobs.{$action['action']}");
```
maxbrokman, line 18: this is very laravel, but personally I'd have an explicit switch here with some logging for actions we don't understand

app/Http/Controllers/External/VatsimNet/ProcessVatsimNetWebhook.php
CalumTowers: ->defer() actually handles the sync driver proper and waits until after the response is returned too. So probably best of both worlds to do that.

app/Http/Controllers/External/VatsimNet/ProcessVatsimNetWebhook.php
CalumTowers: Good shout.

AxonC (commented):
Feels like some tests would be cracking here 😉

PR #3972, opened 2025-02-25
Auto publish filament assets

maxbrokman:
Admin panel will not render / build without them. Discovered during running Core seminar.


CalumTowers:
```
"post-update-cmd": [
            "@php artisan vendor:publish --force --tag=livewire:assets --ansi",
            "@php artisan filament:upgrade"
        ]

```

PR #3978, opened 2025-03-05
feat: add CTS booking API

kristiankunc:
This PR introduces a GET `/api/cts/bookings?date=YYY-MM-DD` route which returns  a list of CTS bookings on that date.

Minimum allowed date is `now - 30 days`

### DIFF
--- a/app/Http/Controllers/Api/CtsController.php
+++ b/app/Http/Controllers/Api/CtsController.php
@@ -3,46 +3,76 @@
 namespace App\Http\Controllers\Api;
 
 use App\Repositories\Cts\BookingRepository;
+use Carbon\Carbon;
 use Illuminate\Http\Request;
-use Illuminate\Support\Facades\RateLimiter;
 use Illuminate\Http\Response;
+use Illuminate\Support\Facades\RateLimiter;
 
 class CtsController
 {
-    public function __construct(private BookingRepository $bookingRepository)
-    {}
+    public function __construct(private BookingRepository $bookingRepository) {}
 
     public function getBookings(Request $request)
     {
         if (app()->environment() !== 'development' && RateLimiter::tooManyAttempts('get-bookings:'.$request->ip(), 1)) {
             $seconds = RateLimiter::availableIn('get-bookings:'.$request->ip());
+
             return response()->json([
-            'message' => 'Too many requests. Please try again after '.$seconds.' seconds.'
+                'message' => 'Too many requests. Please try again after '.$seconds.' seconds.',
             ], Response::HTTP_TOO_MANY_REQUESTS)
-            ->header('Retry-After', $seconds)
-            ->header('X-RateLimit-Reset', now()->addSeconds($seconds)->getTimestamp());
+                ->header('Retry-After', $seconds)
+                ->header('X-RateLimit-Reset', now()->addSeconds($seconds)->getTimestamp());
         }
 
         RateLimiter::hit('get-bookings:'.$request->ip(), 300); // 5 minutes
 
-        $bookings = $this->bookingRepository->getBookings(50);
+        $date = $request->get('date', null);
 
-        if ($bookings->isEmpty()) {
-            return response()->json(['message' => 'No bookings found'], 404);
+        // Validate date, default to today
+        if ($date) {
+            try {
+                $date = Carbon::parse($date);
+
+                if ($date->isPast() && $date->diffInDays(Carbon::now()) > 30) {
+                    return response()->json([
+                        'message' => 'Date is too far in the past. Please use a date within the last 30 days. Oldest date allowed: '.Carbon::now()->subDays(30)->toDateString(),
+                    ], Response::HTTP_BAD_REQUEST);
+                }
+
+            } catch (\Carbon\Exceptions\InvalidFormatException $e) {
+                return response()->json([
+                    'message' => 'Invalid date format. Please use YYYY-MM-DD.',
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/CtsController.php
```diff
@@ -0,0 +1,39 @@
...
+use Illuminate\Http\Request;
+
+class CtsController
+{
+    private $bookingRepository;
+
+    public function __construct(BookingRepository $bookingRepository)
+    {
+        $this->bookingRepository = $bookingRepository;
+    }
```
AxonC, line 15: nitpick: you might benefit from the new constructor initialisation syntax

```suggestion
    public function __construct(private BookingRepository $bookingRepository)
    {}
```

routes/api.php
```diff
@@ -2,6 +2,7 @@
 
 Route::get('validations')->uses('Api\ValidationsController@view')->name('api.validations');
 Route::get('metar/{airportIcao}')->uses('Site\MetarController@get')->name('api.metar');
+Route::get('/cts/bookings')->uses('Api\CtsController@getBookings')->name('api.cts.bookings');
```
AxonC, line 5: General feedback here on a public endpoint: we'd like to apply some rate limits here to ensure we don't get undue load on our database.

app/Repositories/Cts/BookingRepository.php
```diff
@@ -8,6 +8,11 @@
 
 class BookingRepository
 {
+    public function getBookings($perPage = 50)
```
AxonC, line 11: A general comment on this architectural decision: I'd like to see the pagination done in the HTTP layer so we can re-use this repository method in future to quite simply get a list of bookings.

app/Http/Controllers/Api/CtsController.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function getBookings(Request $request)
```
AxonC, line 17: To avoid infinite pagination, I feel it wise here to enforce a maximum 'date in the past' functionality here. Loose suggestion would be you can't retrieve more than 30 days in the past.

routes/api.php
AxonC, line 5: Suggestion: I would suggest that the endpoint accepts a date range and we default to the last 7 days worth of bookings by default.

app/Http/Controllers/Api/CtsController.php
```diff
@@ -0,0 +1,39 @@
...
+        if ($bookings->isEmpty()) {
+            return response()->json(['message' => 'No bookings found'], 404);
+        }
+
+        // Not all bookings from the min/max date may be returned if the pagination limit is reached
+        $fromDate = $bookings->first()->date;
+        $toDate = $bookings->last()->date;
+
+        return response()->json([
+            'bookings' => $bookings,
```
AxonC, line 30: I think we'd like to be quite selective with the fields we are surfacing via this API. If you can identify the bare minimum fields, I think that would be useful. For example: `member_id` isn't actually their CID... I can talk more about that via DM if it's useful.

Looking at the schema I'd suggest plucking the following fields:
- id
- date
- from
- to
- position
- type

AxonC (changes requested):

kristiankunc:
No Idea why the test fails, I've even reverted the `getTodaysBookings()` back to main

AxonC:
> No Idea why the test fails, I've even reverted the `getTodaysBookings()` back to main

`) Tests\Unit\CTS\BookingsRepositoryTest::it_can_return_a_list_of_todays_bookings_with_owner_and_type
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'id' => 96155
-    'date' => Illuminate\Support\Carbon Object (...)
+    'date' => '2025-03-20'
     'from' => '17:00'
     'to' => '20:59'
     'position' => 'EGCC_APP'`

AxonC:
> No Idea why the test fails, I've even reverted the `getTodaysBookings()` back to main

I've just re-ran the tests if you need the logs again

app/Http/Controllers/Api/CtsController.php
```diff
@@ -0,0 +1,80 @@
...
+
+            } catch (\Carbon\Exceptions\InvalidFormatException $e) {
+                return response()->json([
+                    'message' => 'Invalid date format. Please use YYYY-MM-DD.',
+                ], Response::HTTP_BAD_REQUEST);
+            }
+
+        } else {
+            $date = Carbon::now()->startOfDay();
+        }
```
AxonC, line 50: You could simplify this code by defining `$date` above on line 31, and then taking a user requested date accordingly.

app/Http/Controllers/Api/CtsController.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $bookings = $this->bookingRepository->getBookings($date);
+
+        return response()->json([
+            'bookings' => $bookings->map(function ($booking) {
+                return collect($booking)->except(['member'])->toArray();
```
AxonC, line 56: Worth adding a comment as to why we're excluding `member` per my DM?

AxonC (commented):
Couple of minor things before merging - thanks for sticking with this!

app/Http/Controllers/Api/CtsController.php
kristiankunc: Sorry don't remember what the reason was, other than we just won't want to expose as much personal info

AxonC (approved):

PR #3979, opened 2025-03-06
Add a linting pre-commit hook

kristiankunc:
This PR adds a pre-commit hook which will automatically lint all `.php` files staged for the commit.

A setup script is included which when executed will create a symlink from `.git/hooks/` to `git-hooks/` for every hook defined in the directory, meaning the setup script only has to be run for the creating of the hooks, after that they can be versioned normally.

### DIFF
--- a/git-hooks/pre-commit
+++ b/git-hooks/pre-commit
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+echo "Running pre-commit lint hook"
+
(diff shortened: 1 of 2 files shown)


CalumTowers:
Thanks for the PR - we aren't going to merge this right now but we are thinking about how we improve the automation around linting and quote quality more broadly.

PR #3980, opened 2025-03-06
Add development-only docker compose

kristiankunc:
This PR adds a development-only compose config, taken from https://github.com/VATSIM-UK/core/blob/main/.devcontainer/docker-compose.yml but excluding core itself.

Also a utility script in `.devcontainer/load-dotenv.sh` has been added which will load all env variables defined in .env and pass them onto the command executed.

For example:
```sh
.devcontainer/load-dotenv.sh docker-compose -f .devcontainer/docker-compose.dev.yml up
```

### DIFF
--- a/.devcontainer/docker-compose.dev.yml
+++ b/.devcontainer/docker-compose.dev.yml
@@ -1,0 +1,37 @@
+services:
+  redis:
+    image: redis:7-alpine
+    ports:
+      - "6379:6379"
+    command: redis-server --requirepass root
+    environment:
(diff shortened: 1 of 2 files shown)


AxonC:
PR description is really helpful - can we put it somewhere in the devcontainer so its not lost in the merge?

kristiankunc:
I've added an example usage to the top of the compose file, more info is found in the utility script itself.

AxonC (approved):

PR #3996, opened 2025-03-24
feat: remove member search from feedback page

AxonC:


resources/views/mship/feedback/form.blade.php
```diff
@@ -5,106 +5,11 @@
...
-    // Bind input change for serch input
-    searchInput.bind("input", function(event){
-
-        $('#memberSearchSpinner').show();
-        $('#memberSearchNoResults').hide();
-        $('#memberSearchResults').hide();
-
-        clearTimeout(timeout);
-        timeout = window.setTimeout(function(){
-          $.get('{{route('mship.feedback.usersearch', "")}}/' + $('#member-search-name').val(), function(response){
```
atoff, line 23: we should probably remove this route too?

atoff (changes requested):

resources/views/mship/feedback/form.blade.php
AxonC: good spot! fixed

CalumTowers (approved):

atoff (approved):

PR #4029, opened 2025-03-26
TECH-280: Protect New S1 Members from Roster Removal

CalumTowers:
This update will stop newly qualified home S1 controllers from being removed from the roster for not meeting the hour requirements in the first three months of holding their rating.


AxonC (approved):

AxonC:
Wouldn't mind a test for this case 😉

CalumTowers:
> Wouldn't mind a test for this case 😉

In fairness this whole command needs a test!

PR #4045, opened 2025-04-03
Add requires_roster_membership flag to waiting lists to exempt from removal

AxonC:


atoff (commented):
Do we need to add something in filament to allow the editing of this?

AxonC:
> Do we need to add something in filament to allow the editing of this?

It's going to be the minority of lists this applies to so I'm happy with it not being in Filament. Same applies to the feature toggles we have.

PR #4053, opened 2025-04-07
Add facility for waiting lists to allow members to self enrol

AxonC:
Not feature complete for other lists - primarily intended for OBS -> S1 list now in Core.

Still to come:
- hours at rating X post rating
- minimum rating (thinking lists like Heathrow where anyone S1+ can apply to do the Ground rating etc.


app/Services/Training/WaitingListSelfEnrolment.php
```diff
@@ -0,0 +1,58 @@
...
+        Account $account,
+        WaitingList $waitingList
+    ): bool {
+        if ($waitingList->includesAccount($account)) {
+            return false;
+        }
+
+        if (! $waitingList->self_enrolment_enabled) {
+            return false;
+        }
```
CalumTowers, line 23: I would make this the first check. It'll be cheaper than the first query above.

app/Services/Training/WaitingListSelfEnrolment.php
```diff
@@ -0,0 +1,58 @@
...
+
+        if ($waitingList->self_enrolment_maximum_qualification_id) {
+            $requiredQualification = Qualification::find(
+                $waitingList->self_enrolment_maximum_qualification_id
+            );
+
+            if (! $account->hasQualification($requiredQualification)) {
+                return false;
+            }
+        }
```
CalumTowers, line 45:
```suggestion
        if ($waitingList->self_enrolment_maximum_qualification_id && ! $account->hasQualification($waitingList->self_enrolment_maximum_qualification_id)) {
                return false;
        }
```

I'll let you decide if this is better or easier to tidy up some other way. Not a huge fan of multi-level nested if statements.

resources/views/mship/waiting-lists/_waiting_lists_list.blade.php
```diff
@@ -64,6 +64,40 @@
                     @endempty
                 </tbody>
             </table>
+
+            @if ($department === \App\Models\Training\WaitingList::ATC_DEPARTMENT)
+            <div>
+                <h3>Self-enrolment</h3>
+                <p>Some waiting lists have the ability to 'self-enrol' without having to contact the Training team.
+                    If you are eligible to self-enrol, you will see a button below. Please note that this is only available for some waiting lists.</p>
```
CalumTowers, line 72:
```suggestion
                <p>Some waiting lists have the ability to 'self-enrol' without having to contact the Training team.
                    If you are eligible to self-enrol, you will see a button below. </p>
```
Remove repeated "only some lists" as you say it at the start.

app/Services/Training/WaitingListSelfEnrolment.php
AxonC: Mainly there because the min or max qualification is nullable. I'll have a think

PR #4067, opened 2025-04-19
Ops changes

Lucak1011:
- Add AM as ops director
- Add JE as Ops Assistant


Lucak1011:
Not sure why there is so many commits there but they don't seem to have changed any files..
When testing locally, AMs profile picture causes the staff page to load, my impression was that it's his forum ID that it is pulled from but feel free to correct me on that one :)

AxonC (changes requested):
This file needs updating as well for the image to work properly.

https://github.com/VATSIM-UK/core/blob/95f7dc576b31e01e2e188bda4f42ce1942c4f232/app/Http/Controllers/Site/StaffPageController.php#L22-L30

Lucak1011:
That should be done now. Cheers Callum!

AxonC (approved):

AxonC:
> That should be done now. Cheers Callum!

Thanks for the contribution!

PR #4068, opened 2025-04-21
Add ability to toggle self-enrolment for a minimum number of networkdata atc hours at given qualification

AxonC:
Flirted with the idea of having an index over `account_id,qualification_id` but I don't think the volume of data is sufficient.
Equally flirted with having the 'hours' column as a 'minutes' column to avoid some floating point stuff but didn't think it added much really.
If anyon has a better idea for those column names I'm all ears...


CalumTowers (approved):
Nice work.

PR #4072, opened 2025-04-21
refactor: migrate Heathrow training page to markdown

kristiankunc:
Created 2 components containing Heathrow page contents + faqs. Tried playing around with the images as well and determined they'll have to be included as html to make them not massive but also allow responsive scaling.


AxonC (approved):
Great work! Looks so much cleaner.

PR #4094, opened 2025-05-11
feat: Waiting list removal modal

WillShaw1100:
Logs the staff member who performed the action and the reason when removing members from a waiting list.
Ref: [Waiting List Dev Guide](https://github.com/VATSIM-UK/core/blob/b689c6ce5efeeb4235af466f1f13291c9fec77e3/docs/waitinglists/index.md)
image
image

### DIFF
--- a/app/Filament/Resources/WaitingListResource/RelationManagers/AccountsRelationManager.php
+++ b/app/Filament/Resources/WaitingListResource/RelationManagers/AccountsRelationManager.php
@@ -2,6 +2,7 @@
 
 namespace App\Filament\Resources\WaitingListResource\RelationManagers;
 
+use App\Models\Mship\Note\Type;
 use App\Models\Training\WaitingList;
 use App\Models\Training\WaitingList\WaitingListAccount;
 use AxonC\FilamentCopyablePlaceholder\Forms\Components\CopyablePlaceholder;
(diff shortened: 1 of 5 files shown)


database/migrations/2025_05_11_030824_add_removal_reason_and_deleted_by_to_waiting_list_account.php
```diff
@@ -0,0 +1,30 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('training_waiting_list_account', function (Blueprint $table) {
+            $table->unsignedInteger('removed_by')->nullable();
+            $table->text('removal_reason')->nullable();
```
CalumTowers, line 16: We should be specific about which columns these are going after by using ->after()

app/Filament/Resources/WaitingListResource/RelationManagers/AccountsRelationManager.php
```diff
@@ -117,11 +117,43 @@ public function table(Table $table): Table
...
+                            ->reactive(),
+
+                        Forms\Components\Textarea::make('custom_reason')
+                            ->label('Custom reason')
+                            ->rows(3)
+                            ->required()
+                            ->visible(fn (callable $get) => $get('reason') === 'other'),
+                    ])
+                    ->action(function (WaitingListAccount $record, array $data, $livewire) {
+                        $removal_reason = $data['reason'] === 'other' ? $data['custom_reason'] : $data['reason']; // if its other set the $removal_reason as the custom value (instead of 'other (specify below)')
```
CalumTowers, line 140: This perhaps highlights the need to distinguish the reason vs the comment. 

removed_by
removal_type - this effectively becomes an enum
removal_comment

You may not need the comment to be attached to the record in the same table, it could instead be a note on the account which I believe should be polymorphic.

WillShaw1100:
Created a new note type "Training" as I'm thinking this can be used elsewhere in future.

This now also only saves the type in the training_waiting_list_account table for potential stats purposes. The reason is saved as a note on the members account.

maxbrokman:
Continues with fixes in #4096

maxbrokman:
Resolved by #4096!

Thanks for this @WillShaw1100

PR #4097, opened 2025-05-15
feat: Waiting List Create & Flags Changes

WillShaw1100:
QoL changes to Administrators creating waiting lists by providing the options front-end.
Upon creating a waiting list flag, allow the option of displaying it in the table.

Previews:


WillShaw1100:
I have tried to implement some learnings from last time, I'm unsure if we really need to update tests / create them for this as a lot of it is visual and QoL changes, if you want some let me know and I can take a look further...

AxonC:
I think it would be wise to hide the ability to toggle these options behind the waiting list admin permissons

WillShaw1100:
> I think it would be wise to hide the ability to toggle these options behind the waiting list admin permissons

They appear in the create waiting list and create flag components so would already be protected by those permissions, unless i'm mistaken?

WillShaw1100:
> @WillShaw1100 merge conflicts :(

I know the pain... Thankfully this was a quick and easy one 👍

PR #4165, opened 2025-06-23
Fixes #4163 - Replace all instances of factory(WaitingList::class)->create() with new factory format

### LINKED ISSUE #4163: Replace all instances of `factory(WaitingList::class)->create()` with new factory format.
Waiting list tests were written in an age where the factory syntax was different within the Laravel framework. You can see [here](https://github.com/VATSIM-UK/core/blob/main/database/factories/WaitingListFactory.php). The newer syntax is along the lines of `WaitingList::factory()->create()` where the `HasFactory` trait is defined on the model.

This issue should address the inconsistency in the not-so-few waiting list factory invocations.

CLC0609:
Fixes #4163 

# Summary of changes
Replaced all instances of ```factory(WaitingList::class)->create()``` with ```WaitingList::factory()->create()```

# Screenshots (if necessary)

### DIFF
--- a/tests/Feature/Account/WaitingListsTest.php
+++ b/tests/Feature/Account/WaitingListsTest.php
@@ -29,7 +29,7 @@ class WaitingListsTest extends TestCase
     #[Test]
     public function test_index_with_no_waiting_list_accounts()
     {
-        factory(WaitingList::class)->create(['name' => 'My List']);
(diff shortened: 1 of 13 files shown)


CLC0609:
Do the instances of ```factory(WaitingListFlag::class)->create();``` also need to be changed?

PR #4172, opened 2025-06-24
WIP: Booking System

MrAdder:
# Summary of changes
Heavily a WIP
A booking system that pulls from the current CTS (dev environment) to display only.

# Screenshots (if necessary)

I95ijBPl2K

I0ue83T2tv

- [ ] member_id to name 
- [ ] set Zulu time
- [ ] Auto remove past current Zulu time


kristiankunc (changes requested):
1) When the booking in question is at the bottom of the page, the popover stays mostly hidden on hover

Image

2) When I click on a booking, I can then hover over others which overlays the popovers:

Image

3) Display old bookings seems to be buggy:
 - Old bookings hidden
Image
- Old bookings shown
Image

(note that the date is question is in the future so the button should not have any effect on them

4) The booking JSON api exposes names of trainees for mentoring sessions, I'd honestly just scrap the API, the already is one at /api/cts/bookings

5) We should probably stick to the CTS naming eg.
 - Sessions: FirstName FirstSurnameLetter (cid)
 - Exams/Mentoring: show mentor/examiner full name + (cid)

6) Now that the calendar is in core, I think we could implement permissions where staff members can see names on all bookings while normal users can't

MrAdder:
@kristiankunc 1 to 3 is done I will double check

4) Adding a system in that will prevent that as eventually want to pull this fully into Core rather than the CTS DB however it will implement 6

5) to be fair when I did it didnt think of naming Conventions

kristiankunc:
Any updates on this?

kristiankunc:
This has been open for ages and we're discussing how exactly do we want the calendar implemented, if you'd be interested to be a part of it, let me know but closing for now.

PR #4180, opened 2025-06-29
Add capacity limits to training waiting lists

CalumTowers:
Introduces a max_capacity field to training waiting lists, including migration, model logic, admin UI, and enforcement in self-enrolment and manual addition. Adds comprehensive feature and unit tests to verify capacity enforcement, admin management, and self-enrolment behavior.

Fixes TECH-379.


AxonC:
Would the self-enrol button on the Mship waiting lists page be hidden when the capacity limit has been reached?

PR #4184, opened 2025-06-30
Fixes #4178 - Add feedback form i.e. feedback type to feedback admin panel

### LINKED ISSUE #4178: Add feedback form i.e. feedback type to feedback admin panel
For those who can see more than one type of feedback at present they cannot see the feedback type until they press 'View' on the feedback. The table columns should be amended in order to display the feedback form.

Source: https://github.com/VATSIM-UK/core/blob/main/app/Filament/Resources/FeedbackResource.php

CLC0609:
Fixes #4178

# Summary of changes
Added ```Feedback Type``` column
Made ```Feedback Type``` column sortable

# Screenshots (if necessary)
image


AxonC (approved):
Thanks!

PR #4187, opened 2025-06-30
Fixes #4186 - Add ability to copy email from account in admin panel

### LINKED ISSUE #4186: Add ability to copy email from account in admin panel
Small QoL feature to be able to copy the email address in the admin panel. Only for the primary email; form field should remain readonly.

CLC0609:
Fixes #4186

# Summary of changes
Made ```Primary Email``` a ```CopyablePlaceholder```, this will mean it is no longer a ```form field```, this means it is read only. Let me know if the ```Primary Email``` has to remain as a text input specifically. I've made the formatting match the copying option from the ```CID``` field at the top of the same page for consistency.

# Screenshots (if necessary)
image

### DIFF
--- a/app/Filament/Resources/AccountResource.php
+++ b/app/Filament/Resources/AccountResource.php
@@ -80,16 +80,15 @@ class AccountResource extends Resource implements DefinesGatedAttributes
 
                     Forms\Components\Fieldset::make('Emails')->schema([
                         Forms\Components\TextInput::make('email')
-            ->label('Primary Email')
-            ->disabled()
-            ->suffixAction(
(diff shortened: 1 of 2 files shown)


CLC0609:
The tests require ```Primary Email``` to be a ```text input```, because of this I've added a button that allows the user to copy the email while the ```text input``` still exists and is disabled. It's no longer consistent with the ```CID``` field.
image

There is also a tooltip added when the user hovers over the button, let me know if this is wanted or not
image

AxonC:
I prefer the second solution - thanks again, great work!

PR #4194, opened 2025-07-07
Fixes #4193 - Instant search result navigation for account search

### LINKED ISSUE #4193: Instant search result navigation for account search
Within the admin panel, when searching with a CID and there is one result, navigate straight to that member.

CLC0609:
Fixes #4193

# Summary of changes
When searching with a CID on the accounts page of the admin panel, when there is one result it navigates you straight to that member.

# Screenshots (if necessary)

https://github.com/user-attachments/assets/d3be5ce9-dc2d-4d9c-b17f-ddd3643cd762

Wasn't sure how to show this as a screenshot so I figured a short video would be best.

### DIFF
--- a/app/Filament/Admin/Resources/AccountResource/Pages/ListAccounts.php
+++ b/app/Filament/Admin/Resources/AccountResource/Pages/ListAccounts.php
@@ -4,6 +4,7 @@ namespace App\Filament\Admin\Resources\AccountResource\Pages;
 
 use App\Filament\Admin\Helpers\Pages\BaseListRecordsPage;
 use App\Filament\Admin\Resources\AccountResource;
+use App\Models\Mship\Account;
 
 class ListAccounts extends BaseListRecordsPage
 {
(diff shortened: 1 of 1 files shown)


AxonC:
Thanks!  Couple of questions
- Can we also have this on the main search bar at the top?
- Is there any debounce on this?
- What happens if you copy/paste a CID into the box?

CLC0609:
- I will look into adding this onto the global search now
- I don't believe there is a debounce on this, is this needed?
- Copying and pasting a CID into the box redirects the user to the matching account page as expected

CLC0609:
Hi @AxonC , I've been looking into this for the last week or so and really can't find a way to make this work for global search. I'm still very inexperienced in Laravel so there's a good chance I've missed something. Should I let someone else handle this issue who knows how to make these changes?

AxonC (approved):
Thanks this is still an improvement!

PR #4199, opened 2025-07-08
Migrate waiting list retention checks to core

kristiankunc:
Fixes TECH-401

# Summary of changes

Migrate waiting list retention checks to core

# Screenshots (if necessary)
image
image


app/Http/Controllers/Mship/WaitingLists.php
```diff
@@ -47,4 +48,24 @@ public function selfEnrol(WaitingList $waitingList, Request $request)
...
+        $retentionCheck = WaitingListRetentionChecks::where('token', $token)->first();
+
+        // Only the scheduled command will change the status so we need to check the expires_at timestamp as well
+        if ($retentionCheck == null || $retentionCheck->status !== WaitingListRetentionChecks::STATUS_PENDING || $retentionCheck->expires_at < now()) {
+            return abort(404, 'Invalid or expired code');
+        }
+
+        $retentionCheck->response_at = now();
+        $retentionCheck->status = WaitingListRetentionChecks::STATUS_USED;
+        $retentionCheck->save();
```
AxonC, line 69: What sort of success do we show to the user once they have clicked on the token? They might not be logged in at the time but we'd benefit from some form of flash message (not sure I have the solution here since that does require auth), or an email confirmation to say they have had their check saved. Otherwise they have no feedback

resources/views/emails/training/waiting_list_failed_retention.blade.php
```diff
@@ -0,0 +1,11 @@
+@extends('emails.messages.post')
+
+@section('body')
+
+    <p>On {{ $removal_date }} you have been removed from {{ $waiting_list_name }} due to failing to engage in a quarterly
```
AxonC, line 5: Have you got sign-off on the copy text from Adam/Will?

kristiankunc: resolved

PR #4204, opened 2025-07-18
Fix visit-transfer 419 error on application start

kristiankunc:
Fixes *nil*

# Summary of changes

Add a missing csrf token to the form


kristiankunc:
yeah no I messed up here

PR #4246, opened 2025-08-04
Fixes #4245 - Puts All Policies In Dropdown Navigation In Same List

### LINKED ISSUE #4245: Remove 'Web Policy' from policy nav dropdown and unify options vs guidelines
- Remove the 'Web Policy' from navigation dropdown as a divider and instead just have a flat list of all policy links

Example below but not just the ones highlighted

Image

polasais:
Removed the two <*li> from the page, so that the policies are one flat list.

Fixes #4245 

# Summary of changes

Removes <*li> for the divider as well as the Web Policies header, so that all policies are in the same list.

# Screenshots (if necessary)

< Please provide screenshots if this will facilitate a faster review of these changes>


kristiankunc:
Thanks for this

PR #4373, opened 2025-10-23
Fixes #4368 - Convert S1 Training Syllabus and Lesson Plans to markdown in Core

### LINKED ISSUE #4368: Convert S1 Training Syllabus and Lesson Plans to markdown in Core
https://community.vatsim.uk/files/downloads/file/222-s1-syllabus-and-lesson-plans/

Convert to markdown. Add to new page. Add under new navigation category "ATC Training Process" on the navigation bar. 

Example: https://github.com/VATSIM-UK/core/blob/main/resources/views/site/policy/atc-training-body.blade.php

Liaely:
Fixes #4368 

# Summary of changes

Converts the S1 training syllabus and lesson plans into markdown, and implements into core.

### DIFF
--- a/tests/Feature/Site/PolicyPagesTest.php
+++ b/tests/Feature/Site/PolicyPagesTest.php
@@ -58,6 +58,6 @@ class PolicyPagesTest extends TestCase
     #[Test]
     public function test_it_loads_the_s1_syllabus()
     {
-        $this->get(route('site.policy.training-process.s1-syllabus'))->assertOk();
+        $this->get(route('site.policy.training.s1-syllabus'))->assertOk();
     }
 }


Liaely:
Ok, at this point I don't really know what's going wrong beyond the result from the workflow, and I don't know how to fix it, so any help would be greatly appreciated.

resources/views/site/training/s1-syllabus-body.php
CLC0609: I think the reason the markdown isn't rendering is because these need to be `.blade.php` files.

AxonC:
> > Tables are broken
> 
> I can't see what's wrong with that table, I know the other ones feature `</br>` elements, but as far as I can tell the first table is syntactically correct. Perhaps I should just re-write it without tables then if it doesn't seem to parse tables correctly.

It it helps you could also revert to raw HTML tables and it would still parse.

kristiankunc:
We use https://github.com/spatie/laravel-markdown which is essentially just a wrapper for CommonMark which supports some extensions (https://commonmark.thephpleague.com/2.4/customization/rendering/) and tables seem to be there so probably try to play around with that to start off with.

Liaely:
I just took a look though the [CommonMark Spec (0.31.2)](https://spec.commonmark.org/0.31.2) and it doesn't seem to specify tables in Markdown anywhere, so I've converted the tables to HTML.

kristiankunc:
Yeah because it requires the table extension - https://commonmark.thephpleague.com/2.7/extensions/tables/

I am fine with html tables just wanna see if we should just extend the markdown functionality further if more docs are gonna be migrated here

Liaely:
> @Liaely I'll add support for commonmark tables, do you need any more features while I'm at it?

Don't think so, thanks!

kristiankunc:
@Liaely tables now supported. you can just cherry-pick and should be good to go with tables - 785f5d8b0c3f3b7703b4598b734da483c617c563

kristiankunc (approved):
happy with all, thanks!

routes/web-public.php
```diff
@@ -63,5 +63,6 @@
         Route::get('/data-protection-policy')->uses('PolicyPagesController@viewDPP')->name('data-protection');
         Route::get('/branding-guidelines')->uses('PolicyPagesController@viewBranding')->name('branding');
         Route::get('/streaming-guidelines')->uses('PolicyPagesController@viewStreaming')->name('streaming');
+        Route::get('/training-process/s1-syllabus')->uses('PolicyPagesController@viewS1Syllabus')->name('s1-syllabus');
```
kristiankunc: This is probably what you were aiming for
```suggestion
        Route::group([
            'as' => 'training-process.',
            'prefix' => 'training-process',
        ], function () {
            Route::get('/s1-syllabus')->uses('PolicyPagesController@viewS1Syllabus')->name('s1-syllabus');
        });
```

PR #4379, opened 2025-10-25
feat: View Theory exam results in training panel

WillShaw1100:
# Summary of changes

This pull request adds theory exam results into core, in a similar fashion to how practical results have been implemented.
It is step 1 of moving theory exam management over to core with question management, results overrides etc coming later. 

It contains 6 permission nodes for each level:
```
'training.theory.access',
'training.theory.view.*',
'training.theory.view.obs',
'training.theory.view.twr',
'training.theory.view.app',
'training.theory.view.ctr',
```
It also contains 17 tests, testing filters, results visibility and question visibility.

# Screenshots
image

image

image


WillShaw1100:
image

image


Its probably worth also updating the datetime in practical exams history,

PR #4400, opened 2025-11-02
Dependabot refactor

kristiankunc:
- Bi-weekly runs
 - Cleaned up old version ignores

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,5 +23,6 @@ updates:
   - package-ecosystem: composer
     directory: /
     schedule:
-      interval: monthly
+      interval: cron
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -3,31 +3,15 @@ updates:
 - package-ecosystem: github-actions
   directory: "/"
   schedule:
-    interval: weekly
-    day: friday
-    time: "18:00"
+    interval: monthly
```
AxonC: Can we do bi-weekly runs?

kristiankunc: Sources seem to suggest it is possible
https://github.com/dependabot/dependabot-core/issues/12246

PR #4443, opened 2025-12-16
fix: Replace AFV with Track Audio on New Controller page

CLC0609:
Fixes #4441 

# Summary of changes
- Changed AFV to Track Audio, with a link to the Github releases, in the FAQs section. Track Audio is now required over AFV as of #4170

# Screenshots (if necessary)
image

PR #4449, opened 2025-12-16
Revert "fix: Replace AFV with Track Audio on New Controller page"

kristiankunc:
Reverts VATSIM-UK/core#4443


kristiankunc:
@CLC0609 Messed up the PRs, I'll revert yours back and go forward with #4447 instead

PR #4450, opened 2025-12-23
Update TFP page

aidenbok203:


resources/views/site/pilots/tfp.blade.php
```diff
@@ -253,7 +228,7 @@
 
                                 <ul>
                                     <li style="margin-left:18pt;">
-                                        At least 150 hours of flying on the VATSIM network;
+                                        At least 200 hours of flying on the VATSIM network, with the exception of real life pilots (considered on a per-case basis);
```
kristiankunc: Maybe a better wording here like "with the exception of pilots holding a RW license"?

PR #4461, opened 2026-01-02
Add exam forward action to training place view

kristiankunc:
# Summary of changes

Big forward button which forwards a member to a practical exam (same logic as exam setup)
Added a small checkbox indicating the the member already has a pending exam booking (which disables the forward button)

_modal for forwarding, position is set to default but is editable_
image

_exam booking already exist_
image


kristiankunc:
switched to #4463

PR #4482, opened 2026-01-12
Fixes #4458 - Improve autosave behaviour in exam report

### LINKED ISSUE #4458: Improve autosave behaviour in exam report
At the moment the exam report autosave can be a little aggressive. Explore general ways to making it slightly more robust, like a larger debounce time etc.

Success is measured ultimately by autosave still being useful but not disruptive like some examiners have fed back.

CLC0609:
Fixes #4458

When I was testing the exam report autosave I noticed it was mainly disruptive due to the high number of notifications that were sent when autosaving happened. I thought we needed a better way to show to the examiner that the report was saved without spamming their screen whenever they stopped typing.

# Summary of changes
- The save button now changes states between `saved` and `save` to show to the examiner if their report is saved without the need of notifications (the button works the same regardless of state)
image
image

- A smaller debounce time of 500ms detects change while the user is typing, changing the button state to `save` and telling the autosave function that a save is needed
- New variables `autosaveIdleSeconds` the amount of time after the user is idle before an autosave, and `autosaveMinInterval` the minimum amount of time between autosave throttle to reduce constant saving. They also allow easy modifications in the future
- When the user has not been typing for one second and it is over 5 seconds from the last autosave, the system will save the report (without notification) and change the save button state to `saved`. 

 To me personally this felt significantly less intrusive than the previous system as there is no constant notifications, however the report still saves regularly reducing the risk of losing work, and still allows the examiner to tell if their report has been saved or not.

*not 100% if this is exactly how you wanted this change to be made but it was the best option I came up with*

# Screenshots (if necessary)

### DIFF
--- a/resources/views/filament/training/pages/conduct-exam.blade.php
+++ b/resources/views/filament/training/pages/conduct-exam.blade.php
@@ -1,5 +1,5 @@
 <x-filament-panels::page>
-    <div wire:poll.1s="autosave"></div>
+    <div wire:poll.5s="autosave"></div>
     {{ $this->examDetailsInfoList }}
 
     {{ $this->form }}


resources/views/filament/training/pages/conduct-exam.blade.php
```diff
@@ -1,4 +1,5 @@
 <x-filament-panels::page>
+    <div wire:poll.1s="autosave"></div>
```
AxonC: Can you explain the effect of this?

CLC0609: This will call the `autosave` function every 1 second while the user is on the page. 

From what I've read polling can be quite resource intensive if a lot of users are on the page at the same time, however I don't see a huge number of examiners writing all reports all at the same time and "to further cut down on server requests, Livewire automatically throttles polling when a page is in the background. For example, if a user keeps a page open in a different browser tab, Livewire will reduce the number of polling requests by 95% until the user revisits the tab."

This is the best way I could come up with to repetitively check the state of saved state of the report without relying on debounces in the text boxes but there might be a better option than this

AxonC: Interesting all round! Can we perhaps up the frequency to something like 5 seconds just so it isn't quite as aggressive?

AxonC (approved):

PR #4484, opened 2026-01-17
feat: Implement VT Admin into filament panel

WillShaw1100:
# Summary of changes
Implements the VT system into our filament admin panel, with new permissions and tests.

No additional features have been added, for step one it simply moves the existing features over.

I have not included references in the view page, as they are no longer utilised and could possible be deprecated, though of course if you think keeping hold of this is useful they can be easily added.

The old adm-legacy implementation has not been removed and can be pulled out in the future.

Some Screenshots:
image

image


WillShaw1100:
Happy to contribute whenever I can, I actually rather enjoy it.
I have a few ideas for additional features now that its in Core, but need to speak with the rest of the team first.

PR #4486, opened 2026-01-20
feat: Add ability to reallocate feedback to a different member

CLC0609:
# Summary of changes
Added a new button to the header of the view feedback page that allows someone with the `actionFeedback` permission to reallocate the feedback to a different member,

The member search handles both Member Name and CID.

# Screenshots (if necessary)
image
image

(wasn't too sure about the colour of the button but we are running out of colour options for these header actions)


kristiankunc (commented):
A simple test would be nice

PR #4383, opened 2025-10-27
Fixes #4367 - Allow examiners in the exam history section of the training panel to see their past exams

CLC0609:
Fixes #4367 

# Summary of changes
Allows examiners to see exams they conducted even if they no longer have that level of permissions anymore.

# Screenshots (if necessary)

PR #4513, opened 2026-02-02
feat: Allow Examiners to Filter and View Their Conducted Exams

### LINKED ISSUE #4367: Allow examiners in the exam history section of the training panel to see their past exams
Allow examiners who have permission to view exam history to see their past exams they have conducted. This should still respect their current permission sets e.g. if an examiner is no longer an examiner at that level the filter shouldn't remove the permission requirement already in place for the permission assigned.

MrAdder:
Fixes #4367 

### Feature: Allow Examiners to Filter and View Their Conducted Exams

**Related Issue:** #4367
**Related to comments in:** #4383 

#### Changes Made:

1. **app/Filament/Training/Pages/Exam/ExamHistory.php**
   - Added new optional filter "Conducted by me"
   - When checked, filters the exam history table to show only exams where the authenticated user was a conductor (primary, secondary, or trainee examiner)
   - The filter respects the user's current permissions and exam access levels
   - Implemented as a checkbox filter that modifies the query using `whereHas` relationships with the `PracticalExaminers` model

#### How It Works:

- Users with `training.exams.access` permission can access the exam history page
- By default, the table shows exams based on their current conduct permissions
- When the "Conducted by me" filter is enabled, the table filters to show only exams where:
  - The user was the primary examiner (senior), OR
  - The user was the secondary examiner (other), OR
  - The user was the trainee examiner
- The filter matches the user's CID against the examiner records in the `PracticalExaminers` table
- Users can still see other exams without the filter preventing them from doing so

### DIFF
--- a/app/Observers/Training/TrainingPlaceObserver.php
+++ b/app/Observers/Training/TrainingPlaceObserver.php
@@ -32,7 +32,7 @@ class TrainingPlaceObserver
      */
     public function deleted(TrainingPlace $trainingPlace): void
     {
-        //
+        $this->trainingPlaceService->revokeMentoringPermissions($trainingPlace);
     }
 
     /**
--- a/app/Services/Training/TrainingPlaceService.php
+++ b/app/Services/Training/TrainingPlaceService.php
@@ -25,6 +25,22 @@ class TrainingPlaceService
         foreach ($ctsPositions as $ctsPosition) {
             $ctsPositionModel = Position::where('callsign', $ctsPosition)->first();
 
+            if (! $ctsPositionModel) {
+                Log::error("CTS position with callsign {$ctsPosition} not found");
+
+                continue;
+            }
+
+            // Check if the validation already exists to prevent duplicates
+            $existingValidation = PositionValidation::where('member_id', $student->member->id)
+                ->where('position_id', $ctsPositionModel->id)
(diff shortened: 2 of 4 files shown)


app/Filament/Training/Pages/Exam/ExamHistory.php
```diff
@@ -100,6 +100,24 @@ public function table(Table $table): Table
...
+                    if ($data['conducted_by_me']) {
+                        $userCid = auth()->user()->id;
+
+                        return $query->whereHas('examBooking.examiners', function ($q) use ($userCid) {
+                            $q->where(function ($subQuery) use ($userCid) {
+                                $subQuery->whereHas('primaryExaminer', fn ($sq) => $sq->where('cid', $userCid))
+                                    ->orWhereHas('secondaryExaminer', fn ($sq) => $sq->where('cid', $userCid))
+                                    ->orWhereHas('traineeExaminer', fn ($sq) => $sq->where('cid', $userCid));
+                            });
+                        });
```
AxonC: We're trying to move a lot more business logic out of the UI components so that it can be more easily tested. Something like the classes found in 'Services' directory of App.

AxonC (approved):
Great stuff! Would've been happy to keep stuff like badge colour etc in the component but the query logic in the service was exactly what I was after. Thanks!

PR #4517, opened 2026-02-03
feat: Add the ability to set minimum and maximum ratings to VT facilities

CLC0609:
# Summary of changes
- Added `minimum_atc_qualification_id`, `maximum_atc_qualification_id`, `minimum_pilot_qualification_id` and `maximum_pilot_qualification_id`
- Added Select form fields to the Facility resource to allow a admin to select a minimum and maximum. Which pair of columns show depends on the `training_team` of the facility (atc or pilot).
- Added the minimum and maximum ratings onto the public facing facility select, and disabled the apply button if the user does not meet the requirements

I wasn't sure about using one `minimum_qualification_id` and `maximum_qualification_id` column instead of splitting them into two separate pairs of columns. The two pairs seemed like a simpler solution but can change to a single pair if this is preferred.

# Screenshots (if necessary)

### ATC
image

### Pilot
image

---

### ATC
image

### Pilot
image

*requested by WS as a solution to having members without qualifications applying to vist/transfer causing an error during the 90 day check*

### DIFF
--- a/app/Models/VisitTransfer/Application.php
+++ b/app/Models/VisitTransfer/Application.php
@@ -580,7 +580,7 @@ class Application extends Model
 
         $this->guardAgainstApplyingToAFacilityWithNoCapacity($facility);
 
-        if (!$this->meetsRatingRequirements($requestedFacility)) {
+        if (! $this->meetsRatingRequirements($requestedFacility)) {
             throw new RatingRequirementNotMetException($requestedFacility);
         }
 


app/Exceptions/VisitTransfer/Application/RatingRequirementNotMetException.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace App\Exceptions\VisitTransfer\Application;
+
+use App\Models\VisitTransfer\Facility;
+
+class RatingRequirementNotMetException extends \Exception
```
AxonC: Nit: you should be able to import Exception rather than using the preceding slash.

AxonC (commented):
Couple of small bits but looks great to me! Nice work

PR #4525, opened 2026-02-15
Add process for automatically checking availability when training place active

AxonC:


app/Jobs/Training/CheckAvailability.php
```diff
@@ -0,0 +1,113 @@
...
+
+        // Check if availability exists for the student
+        $availability = Availability::where('student_id', $memberId)->get();
+        $hasAvailability = $availability->count() > 0;
+
+        // Check if a session request exists for the training position
+        $hasSessionRequest = $this->checkSessionRequest($memberId);
+
+        // Check passes only if BOTH availability and session request exist
+        if ($hasAvailability && $hasSessionRequest) {
```
CLC0609: I may be wrong but shouldn't there be functionality here to set the previous `AvailabilityCheck` to be `resolved`, I can only see `markWarningAsResolved` being used in the tests currently.

PR #4534, opened 2026-02-15
feat: add delegated role administration

CLC0609:
# Summary of Changes

## Role Resource

A new `Delegates` relation manager has been added to the Role resource.

When no role-specific permission exists for a given role, the interface displays the following:

image

After creating the required permission:

image

Users can then be assigned permission to manage that role:

image

Currently only users with the `admin.access` permission can be delegated the permission to manage a role.

image

Since this is implemented as a standard permission, it can also be assigned to roles once created:

image

---

## RoleRelationManager

On the Accounts page:

- Users with the `account.edit-roles.*` permission can view, attach, and detach **any** roles.
- Users without this permission can only view roles for which they have been explicitly delegated access.
- Users can only attach permissions they have been delegated to manage 

image


AxonC (approved):
Awesome work!

PR #4546, opened 2026-02-21
fix: resolve issue with roster stats not updating

CLC0609:
# Summary of changes
- fix roster stats not showing (Will only fix for future roster updates).

# Screenshots (if necessary)
image


AxonC:
What was the root cause of this out of interest?

PR #4564, opened 2026-03-01
feat: automatically remove training place when an exam is passed

CLC0609:
# Summary of changes
- Automatically remove training place when exam is passed

# Screenshots (if necessary)


AxonC (approved):
Another way to do this could have been via an event but this works for me!

PR #4649, opened 2026-04-05
feat: add pilot examiner management

CLC0609:
# Summary of changes
- Adds the ability to manage pilot examiners
- Adds `training.examiners.view.pilot` and `training.examiners.manage.pilot`
- Adds Pilot Examiner (P?) roles
- Seeds pilot examiner roles based on `ExaminerSettings` cts table

# Screenshots (if necessary)
image


database/migrations/2026_04_08_212122_move_to_new_pilot_examiner_roles_from_cts_examiner_settings.php
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        ExaminerSettings::query()
+            ->where(function ($query) {
+                foreach (array_keys(self::ROLE_MAP) as $column) {
+                    $query->orWhere($column, 1);
+                }
+            })
+            ->each(function (ExaminerSettings $settings) {
+                $account = Account::query()->find($settings->memberID);
```
AxonC: Is this going to use a CTS member ID instead of the account ID?

CLC0609: Yeah I think it would have used the CTS id instead actually. I've updated it to `$settings->member->account->id` and it's working locally

AxonC (approved):

PR #4671, opened 2026-04-09
feat: auto-redirect to record when global search returns single result

CLC0609:
Finishes #4194

# Summary of changes
- auto-redirect to record when global search returns single result when searching by cid

# Screenshots (if necessary)

https://github.com/user-attachments/assets/562f02c8-2c12-457b-b413-f81348f72d57


AxonC (approved):
Is this new with V4? Lovely stuff 😍

CLC0609:
I believe so yeah, or I just couldn't figure it out on V3 when I tried before.

PR #4677, opened 2026-04-10
Backfill account relationship from waiting list account

AxonC:
#### Context
We introduced a nullable `training_places.account_id` FK in PR1 to support a direct relationship from a training place to the student `Account`, while retaining `waiting_list_account_id` for queue provenance and waiting-time stats.

This PR combines the next two rollout stages:
- **PR2**: ensure new training places always write `account_id` (including ad-hoc creation without waiting list records)
- **PR3**: backfill `account_id` for existing `training_places` via the legacy waiting list relationship

#### Summary of changes
- **Dual-write on create**: `TrainingPlaceService::createManualTrainingPlace()` now sets:
  - `waiting_list_account_id` (unchanged)
  - `account_id` from `WaitingListAccount::account_id` (new)
- **Ad-hoc creation**: adds `TrainingPlaceService::createAdhocTrainingPlace(Account $account, TrainingPosition $trainingPosition)`:
  - creates a training place with `account_id` se  - leaves `waiting_list_account_id` as `NULL`
- **Service hardening**: mentoring/exam logic resolves the student via `TrainingPlace::studentAccount()` and fails safely if the account can’t be resolved (prevents crashes once ad-hoc places exist).
- **Backfill migration**: adds a migration to set `training_places.account_id` for existing rows:
  - `training_places.account_id = training_waiting_list_account.account_id`
  - only where `training_places.account_id IS NULL` and `waiting_list_account_id IS NOT NULL`

#### Why this is safe
- Existing behaviour (manual training place creation from waiting list and waiting-list removal) is preserved.
- Backfill only populates missing `account_id`; it won’t overwrite any already-set value.
- Runtime reads are not yet switched globally to `account_id` (that comes in PR4), but the data is now present and new records are written correctly.

#### Testing
- Updated `ManualTrainingPlaceCreationTest` to assert `account_id` is stored.
- Added `AdhocTrainingPlaceCreationTeto verify ad-hoc training place creation without a waiting list account.
- Verified key training-place behaviours remain green by running:
  - `ManualTrainingPlaceCreationTest`
  - `AdhocTrainingPlaceCreationTest`
  - `TrainingPlaceAccountRelationshipTest`
  - `MentoringPermissionAssignmentTest`

### DIFF
--- a/app/Models/Training/TrainingPlace/TrainingPlace.php
+++ b/app/Models/Training/TrainingPlace/TrainingPlace.php
@@ -38,6 +38,20 @@ class TrainingPlace extends Model
             ->withTrashed();
     }
 
+    public function account(): BelongsTo
+    {
+        return $this->belongsTo(Account::class, 'account_id');
+    }
+
+    /**
+     * Transitional accessor to resolve the student account from the new direct FK
+     * (preferred) or the legacy waiting-list relationship (fallback).
+     */
+    public function studentAccount(): ?Account
+    {
+        return $this->account ?? $this->waitingListAccount?->account;
+    }
+
     public function trainingPosition(): BelongsTo
     {
         return $this->belongsTo(TrainingPosition::class, 'training_position_id');
--- a/app/Services/Training/TrainingPlaceService.php
+++ b/app/Services/Training/TrainingPlaceService.php
@@ -6,6 +6,7 @@ use App\Enums\PositionValidationStatusEnum;
 use App\Models\Cts\ExamBooking;
 use App\Models\Cts\Position;
 use App\Models\Cts\PositionValidation;
+use App\Models\Mship\Account;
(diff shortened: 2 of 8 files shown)


CLC0609 (commented):
`ActionExpiredAvailabilityWarningRemoval.php` and`ActionFourthAvailabilityFailureRemoval.php`  still use `$trainingPlace->waitingListAccount->account` should these be moved over to `$trainingPlace->studentAccount();`?

AxonC:
> `ActionExpiredAvailabilityWarningRemoval.php` and`ActionFourthAvailabilityFailureRemoval.php` still use `$trainingPlace->waitingListAccount->account` should these be moved over to `$trainingPlace->studentAccount();`?

Yeah I've got that coming up in my next change

CLC0609 (approved):

PR #4684, opened 2026-04-12
fix: move 'My Training' to top of navigation and reorder exam navigation items

CLC0609:
# Summary of changes
- Reordered navigation groups so My Training is at the top
- Reordered the exam navigation items into a more logical order

# Screenshots (if necessary)
image

### DIFF
--- a/app/Filament/Training/Pages/Exam/ExamHistory.php
+++ b/app/Filament/Training/Pages/Exam/ExamHistory.php
@@ -24,7 +24,7 @@ class ExamHistory extends Page implements HasTable
 
     protected string $view = 'filament.training.pages.exam-history';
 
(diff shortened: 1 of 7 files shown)


kristiankunc (commented):
If we do custom sorting, may be worth leaving some gaps (eg. 10, 20, 30) so if we wanna add some in between, we don't have to update all the pages. But then it's somewhat unpredictable. Probably up to you

CLC0609:
Yeah that makes sense! I've updated to use 10, 20, 30 and 40. I only changed the exam orders because I don't think it makes sense to have "Exam History" above "Exams".

kristiankunc (approved):

PR #4689, opened 2026-04-12
feat: add exam cancellations table to training place page

CLC0609:
# Summary of changes
- Added exam cancellations table to training place page. Exams are linked to training places via the position. So a TP on "EGKK_APP" will show any exam cancellations on "EGKK_APP"
- Fixed some weird null state text on the LOA table

# Screenshots (if necessary)
image

### DIFF
--- a/app/Livewire/Training/TrainingPlaceExamCancellationsTable.php
+++ b/app/Livewire/Training/TrainingPlaceExamCancellationsTable.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace App\Livewire\Training;
+
+use App\Models\Cts\CancelReason;
+use App\Models\Training\TrainingPlace\TrainingPlace;
+use Filament\Actions\Concerns\InteractsWithActions;
+use Filament\Forms\Concerns\InteractsWithForms;
+use Filament\Forms\Contracts\HasForms;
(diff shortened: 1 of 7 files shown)


app/Livewire/Training/TrainingPlaceExamCancellationsTable.php
```diff
@@ -0,0 +1,57 @@
...
+    use InteractsWithForms;
+    use InteractsWithTable;
+
+    public TrainingPlace $trainingPlace;
+
+    public function table(Table $table): Table
+    {
+        $position = $this->trainingPlace->trainingPosition->exam_callsign;
+
+        return $table
```
AxonC, line 28: I wonder whether we don't show this table if there aren't any cancellations to avoid adding extra noise to the page. No problems with what is here, more a musing for the future

AxonC (approved):

app/Livewire/Training/TrainingPlaceExamCancellationsTable.php
CLC0609: Yeah I agree, that page has quite a few tables now 🤣. I'll add something to hide it if there are no cancellations.

AxonC (approved):

PR #4694, opened 2026-04-18
feat: add My Availability to the My Training area

CLC0609:
# Summary of changes
- Adds availability management and validation to the My Training area of the Training Panel
  - Allows a member with `training.access` to view "My Availability"
  - Allows a member to view any future availability
  - Allows a member to edit an availability slot
  - Allows a member to delete an availability slot
  - Allows a member to create a new availability slot
  - Validates any creates or edits for overlapping slots

# Screenshots (if necessary)
image

image

## Changing timezones
- Header actions allows you to choose a time zone from a list
- All times on the page are then converted to that timezone
- Times are saved into the DB as Zulu
- Current timezone selected is shown numerous times across the page
- Users selected time zone is persisted in the session

image

image

- Default time zone is detected by the browser
- The time zone selector list has Zulu and your local at the top, but allows you to select any time zone

image


kristiankunc:
Let's PLEASE allow adding availability in local timezone. I know it will likely get annoying but I'd love it very much.
Can be under a toggle or something so that you can switch between local/zulu

Open to discussion on this

Edit: might be enough to keep inputs in Z but display local time next to it or something like that

CLC0609:
Yep definitely something we need, do you perchance know how the cts knows what your "local" time zone is?

kristiankunc:
CTS probably doesn't but your browser does. Times will have to be stored in Z / whatever CTS is using - I would not change that. The conversion has to happen on the browser.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

CLC0609:
@kristiankunc all implemented.

Any other features you can think of that would be nice to add to this?

kristiankunc:
Looks good - love the timezone feature
A CTS-style daylight savings would be good (can just be hard-coded for the UK dates, let's not bother with the timezones there)

CLC0609:
For daylight savings, if the user uses there local time zone, which in the UK is "London/Europe" then it will show the times in BST.

kristiankunc (approved):
In that case I'm happy. If we get someone feedback on the UI we can always zweak later on. Love this!!

PR #4695, opened 2026-04-18
fix: ensure Tier Endorsement notification fires

CLC0609:
# Summary of changes
- ensure Tier Endorsement notification fires

# Screenshots (if necessary)


kristiankunc (approved):
yeah makes more sense like this rather than having it hidden in Filament

CLC0609:
Yeah it might actually make it fire as well lol 😆

PR #4696, opened 2026-04-18
feat: add banner to alert S1 controllers to the correct Heathrow page

CLC0609:
# Summary of changes
- Added a banner to the top of the Heathrow page only for S1s
- Updated some links in the FAQs section

# Screenshots (if necessary)
image

### DIFF
--- a/app/Http/Controllers/Site/ATCPagesController.php
+++ b/app/Http/Controllers/Site/ATCPagesController.php
@@ -2,8 +2,6 @@
 
 namespace App\Http\Controllers\Site;
 
-use Illuminate\Support\Facades\Redirect;
-
 class ATCPagesController extends \App\Http\Controllers\BaseController
 {
     public function __construct()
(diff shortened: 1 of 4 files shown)


kristiankunc:
Not a fan of forcing it, a notice on the top may be better. This would make the S2 page essentially inaccessible to S1s

CLC0609:
Happy to change it but the S1 page is inaccessible for S2+. Seems to cause some confusion with S1s being allowed access to both.

kristiankunc:
oh I didn't know that, in that case it makes sense to redirect I guess. Still can be kind of confusing that different people can end up on different pages when clicking the same button.

I think we should (instead of redirecting) somehow prompt a redirect to the appropriate page but allow the user to continue if wanted.

AxonC:
Seconded. Never liked the forceful redirect.

AxonC (approved):

PR #4699, opened 2026-04-20
feat: Conditionally append (BST) to "Europe/London" time zone when BST is active

CLC0609:
# Summary of changes
- In order to ensure there is no confusion, during BST "Europe/London" will have a "(BST)" suffix

# Screenshots (if necessary)
image
image


kristiankunc:
let's use something like (not ZULU) to make it clear and to not need to worry about switching it to UTC for every DST

CLC0609:
image
like this?

kristiankunc:
yes, looks good

PR #4704, opened 2026-04-26
fix: ensure training places are removed after exam pass

CLC0609:
# Summary of changes
- Now that waiting list account records are soft deleted following adding a training place, we must also get deleted waiting list accounts

# Screenshots (if necessary)

### DIFF
--- a/app/Filament/Training/Pages/Exam/ConductExam.php
+++ b/app/Filament/Training/Pages/Exam/ConductExam.php
@@ -397,7 +397,7 @@ class ConductExam extends Page implements HasForms, HasInfolists
     public function removeTrainingPlace()
     {
         $studentAccount = $this->examBooking->studentAccount();
-        $waitingListAccountIds = $studentAccount->waitingListAccounts()->pluck('id');
+        $waitingListAccountIds = $studentAccount->waitingListAccounts()->withTrashed()->pluck('id');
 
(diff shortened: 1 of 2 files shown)


AxonC:
We might be able to directly use the account relationship on a training place for this now

CLC0609:
Okay, this should now also not remove both training places if you are training with ATC and Pilot at the same time.

app/Filament/Training/Pages/Exam/ConductExam.php
```diff
@@ -397,9 +397,22 @@ public function markDirty(): void
...
     {
         $studentAccount = $this->examBooking->studentAccount();
-        $waitingListAccountIds = $studentAccount->waitingListAccounts()->pluck('id');
 
-        TrainingPlace::whereIn('waiting_list_account_id', $waitingListAccountIds)->first()?->delete();
+        if (! $studentAccount) {
+            return;
+        }
+
+        $examCallsign = $this->examBooking->position_1;
```
AxonC, line 405: Does this handle the override we have in for positions like OBS_CC_PT3?

CLC0609: I might be missing something, but the exam booking would have a position of "OBS_CC_PT3" because it should be assinging the `exam_callsign` when creating the exam, and then below I am checking both `exam_callsign` and the `position.callsign`.
So I *think* it should handle the override yes.

AxonC (approved):

PR #4712, opened 2026-05-01
Dev env alert

kristiankunc:
Added a small dev env alert to the top panel, same as for UKCP cuz I almost accidentally messed with prod there.

image


kristiankunc:
do we like this?

CLC0609 (commented):
Yep I like it. Could we add it to the training panel aswell?

CLC0609 (approved):

PR #4714, opened 2026-05-02
Hide availability if user does not have a training place

kristiankunc:
Seems useless to have it shown.


CLC0609 (commented):
My only concern would be if we roll out My Availability before we move Pilot Training over to training places. As far as I'm aware pilot training can't yet use training places, so it would mean their students couldn't use the new system.

kristiankunc:
Oh yeah in that case I'll leave it here for now

AxonC:
We could use the PositionValidation table in CTS to check if they have pilot permissions as well as an interim measure.

app/Filament/Training/Pages/MyTraining/MyAvailability.php
```diff
@@ -44,7 +48,43 @@ class MyAvailability extends Page implements HasForms, HasTable
...
+            ->whereHas('position', function ($query) {
+                $query->where(function ($q) {
+                    $q->where('callsign', 'like', 'P1_%')
+                        ->orWhere('callsign', 'like', 'P2_%')
+                        ->orWhere('callsign', 'like', 'P3_%')
+                        ->orWhere('callsign', 'like', 'P4_%')
+                        ->orWhere('callsign', 'like', 'P5_%')
+                        ->orWhere('callsign', 'like', 'P0_%')
+                        ->orWhere('callsign', 'like', 'PT_%')
+                        ->orWhere('callsign', 'like', 'TFP_%');
```
kristiankunc: I think this should cover all of the positions

CLC0609: Would need to check the status to ensure it's not a mentor permission,

The current position validation callsigns for pilot are: (as I just discovered lol)
"P1_PPL(A)"
"P2_SEIR(A)"
"P3_CMEL(A)"

kristiankunc: Mentors should be able to input as well for OTS no?

CLC0609: Ah yeah that's true. Thinking about it you'll need the OTS session permissions as well 

OTS session permission are:
"P1_PPL(A)_MEN"
"P2_SEIR(A)_MEN"
"P3_CMEL(A)_MEN"

i believe

CLC0609: that should allow us to say mentors can't but mentors awaiting OTS' can

kristiankunc: Made it an exact match, should be better

app/Filament/Training/Pages/MyTraining/MyAvailability.php
```diff
@@ -44,7 +48,47 @@ class MyAvailability extends Page implements HasForms, HasTable
...
+        }
+
+        $pilotCallsigns = [
+            'P1_PPL(A)',
+            'P2_SEIR(A)',
+            'P3_CMEL(A)',
+            'P1_PPL(A)_MEN',
+            'P2_SEIR(A)_MEN',
+            'P3_CMEL(A)_MEN',
+        ];
```
CLC0609, line 80:
```suggestion
        $pilotCallsigns = [
            'P1_PPL(A)',
            'P2_SEIR(A)',
            'P3_CMEL(A)',
            'P1_PPL(A)_MEN',
            'P2_SEIR(A)_MEN',
            'P3_CMEL(A)_MEN',
            'TFP_FLIGHT',
        ];
```
I forgot about TFP when I reviewed initially, all looks good other than that.

CLC0609 (approved):

PR #4720, opened 2026-05-03
Bulk feedback processing

kristiankunc:
# Summary of changes

- Subject filter in feedback list
- Bulk actions in feedback list
- Shared action form configs
- Some deprecated method replacement

# Screenshots (if necessary)

image


kristiankunc:
The UI is annoying as hell

CLC0609:
I don't think the UI here is too bad. I mean it's the same number of buttons as on the main feedback page. Even if that is a crazy amount of buttons 😆.

image

Maybe arranging them into an action group as above?

kristiankunc:
yeah but the logic - if 1 feedback is rejected already, the send/action should be probably disabled etc. If they all could work at once that'd be grand

PR #4737, opened 2026-05-04
fix: bump laravel and passport

CLC0609:
# Summary of changes
- bump laravel to 13.7.0 and passport to 13.7.5

# Screenshots (if necessary)


app/Models/Mship/Account/Email.php
```diff
@@ -69,6 +69,11 @@ public function tokens()
         return $this->morphMany(Token::class, 'related');
     }
 
+    public function sysTokens()
+    {
+        return $this->tokens();
+    }
+
```
kristiankunc, line 71: Do we need this? Does not seem to add anything - or is `this.tokens` private?

CLC0609: I seem to remember there being an error somewhere, caused by using tokens, when I did this originally, but I cannot remember. I will take a look tomorrow.

kristiankunc: Alr, I'll hold off on the dependabot PRs so you don't have to start all over again ;)

kristiankunc (approved):

PR #4738, opened 2026-05-04
feat: add a mentoring history page

CLC0609:
# Summary of changes
- Adds a mentoring history page
- You can view the page if you hold ANY mentoring permissions
- You can view TGs you hold ANY mentoring permission within
- You can only see sessions that you hold permission to mentor

# Screenshots (if necessary)
image
image
image
image

Any feedback on the looks / styling of the page is greatly appreciated

### DIFF
--- a/app/Filament/Training/Pages/Mentor/Base/BaseMentoringHistoryPage.php
+++ b/app/Filament/Training/Pages/Mentor/Base/BaseMentoringHistoryPage.php
@@ -88,7 +88,7 @@ abstract class BaseMentoringHistoryPage extends Page implements HasTable
                     ->visible(fn ($record) => $record->filed !== null)
                     ->openUrlInNewTab(),
             ])
-            ->emptyStateHeading('No mentoring sessions found');
+            ->emptyStateHeading('No mentoring sessions found in this training group');
     }
 
     private function getTableFilters(): array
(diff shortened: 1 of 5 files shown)


WillShaw1100:
I imagine you could create the students "My History" page with very little work from this. It will, I assume, contain much of the same info with a slightly altered query. Is there a way to avoid duplication?

It'd also be nice to be able to select multiple status' such as Cancelled and No Show.

CLC0609:
Yep good point, I have extracted the duplicated logic out to a `BaseMentoringHistoryPage` that we can then use to make the student facing session history page.

Good idea, I have implemented the multi-select.

Thanks!

app/Filament/Training/Pages/Mentor/Base/BaseMentoringHistoryPage.php
```diff
@@ -0,0 +1,183 @@
...
+
+    protected function showStudentFilter(): bool
+    {
+        return true;
+    }
+
+    protected function showMentorFilter(): bool
+    {
+        return true;
+    }
```
AxonC, line 37: Is there something to come here that would mean they aren't true?

AxonC: Ahh hang on its the abstract

AxonC (approved):

app/Filament/Training/Pages/Mentor/Base/BaseMentoringHistoryPage.php
CLC0609: My plan is to use that base page for the student facing mentoring history aswell. In which case we wouldn't want the student filter.

PR #4740, opened 2026-05-06
add position table crud

kristiankunc:
image

CRUD on the positions table restricted to OPS for roster changes


app/Filament/Admin/Resources/Positions/Pages/ViewPosition.php
CLC0609, line 1: I don't think this view file is actually needed as we use the modal to edit / view.

CLC0609 (commented):
Looks great!

One comment but other than that all good to go!

CLC0609 (approved):

PR #4744, opened 2026-05-08
feat: Adds membership info to training panel dashboard

WillShaw1100:
# Summary of changes

Adds a bit of purpose to the dashboard

# Screenshots

image


kristiankunc:
I like this. Though I'd remove the roles and just display mentoring positions instead. As a QoL I'd add a copy button to the CID.

Also this could be useful on the admin panel - it could display much more (including the roles like here). If you want feel free to open another PR for it as it looks pretty cool and finally gives some use to the panel homepages

WillShaw1100:
> I like this. Though I'd remove the roles and just display mentoring positions instead. As a QoL I'd add a copy button to the CID.
> 
> Also this could be useful on the admin panel - it could display much more (including the roles like here). If you want feel free to open another PR for it as it looks pretty cool and finally gives some use to the panel homepages

What I will look to do is pass the panel to the view file and use conditional statements to show the positions they can mentor only on the training panel and the roles only on the admin panel. Hopefully that will cut out a lot of duplication.

kristiankunc (commented):
Is it possible to center it vertically?

resources/views/filament/widgets/account-info-widget.blade.php
```diff
@@ -0,0 +1,169 @@
...
+            </div>
+
+            <div class="space-y-3 mt-5">
+
+                {{-- Main grid --}}
+                <div class="grid grid-cols-1 sm:grid-cols-2 gap-x-10 gap-y-8 text-sm">
+
+                    <div x-data="{ copy(text) { navigator.clipboard.writeText(text) } }">
+                        <span class="text-gray-500">CID</span>
+                        <div class="font-medium cursor-pointer hover:text-primary-600 dark:hover:text-primary-400 transition underline underline-offset-4 decoration-dotted" @click="copy('{{ $user->id }}')" title="Click to copy">{{ $user->id }}</div>
```
kristiankunc: This should give some feedback - _cid copied_

WillShaw1100:
> Is it possible to center it vertically?

Im not sure about centering it vertically... maybe I have misunderstood. Using a full column span across looks quite good though.

image

kristiankunc (approved):
Yup, I like that

CLC0609 (approved):
Looks good to me.  Thanks for this!

@AxonC happy with this?

PR #4769, opened 2026-05-16
fix: update training place inactive alert to use a callout

CLC0609:
# Summary of changes
- Use a callout instead of a section because it looks way better

# Screenshots (if necessary)
Before:
image

After:
image


kristiankunc (approved):
I like it

PR #4811, opened 2026-05-24
Allow 'all' option on upcoming mentoring sessions page

AxonC:
## Summary

Adds an **All** option to the training group dropdown on the Upcoming Sessions page, so mentors with access to multiple training groups can view upcoming sessions and pending reports across all of them in one place.

- Shows **All** in the dropdown only when the user can view more than one training group
- Defaults to **All** on first load when multiple groups are available (falls back to the single group when only one is visible)
- Scopes both the **Upcoming Sessions** table and the **Pending Reports** table to all viewable CTS positions when **All** is selected
- Uses `category=all` in the URL to persist the selection

## Changes

- **`UpcomingMentoringSessions`** — adds the **All** dropdown action, default category logic, and position scoping across visible categories
- **`PendingMentoringReportsTable`** — mirrors the same **All** scoping behaviour for the pending reports section on the page
- **`MentorPermissionService`** — adds `ALL_CATEGORIES` constant and `getAllCtsCallsignsFoories()` helper
- **`UpcomingMentoringSessionsTest`** — updates default category assertions and adds coverage for the cross-category **All** view

### DIFF
--- a/app/Filament/Training/Pages/Mentor/UpcomingMentoringSessions.php
+++ b/app/Filament/Training/Pages/Mentor/UpcomingMentoringSessions.php
@@ -46,28 +46,42 @@ class UpcomingMentoringSessions extends BaseMentoringHistoryPage
 
     public function mount(): void
     {
+        if ($this->category === MentorPermissionService::ALL_CATEGORIES) {
+            if (! $this->hasMultipleVisibleCategories()) {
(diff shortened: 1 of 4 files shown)


AxonC:
Feel free to merge once approved.

kristiankunc (approved):

PR #4813, opened 2026-05-24
Show mentoring session indication next to controlling session in training place

kristiankunc:
image

Network session is considered mentoring when it has any overlap with the mentoring session duration

### DIFF
--- a/app/Livewire/Training/RecentControllingTable.php
+++ b/app/Livewire/Training/RecentControllingTable.php
@@ -2,17 +2,21 @@
 
 namespace App\Livewire\Training;
 
+use App\Models\Cts\Session as CtsSession;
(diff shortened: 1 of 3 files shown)


AxonC:
Can you put it next to the callsign instead of adding a whole new column, and then having a tooltip perhaps? Nothing wrong with the current design and happy to iterate but feels like adding a whole new column is a lot of screen space

kristiankunc:
> Can you put it next to the callsign instead of adding a whole new column, and then having a tooltip perhaps? Nothing wrong with the current design and happy to iterate but feels like adding a whole new column is a lot of screen space

fixed - added a tooltip + the icon next to the callsign

AxonC (approved):
love it

PR #4823, opened 2026-05-24
Remember last opened TG across mentoring screens

kristiankunc:
Added a new trait for remembering TG + 2 menthods which manage the sate duging `mount`

@AxonC is this the proper laravel way to do this?


AxonC (approved):
LGTM

PR #4835, opened 2026-05-28
fix: ensure rich text boxes are styled on conduct mentoring pages

CLC0609:
# Summary of changes
- Fix issues with styling of text boxes
- Add a colour selector (I'm not super happy with how long it takes to select a colour, but this is how filament does it natively so it's fine for now, something to look into in the future).
- Add a way to add text to links, and edit them from a click action

# Screenshots (if necessary)
image
image

image
image


tests/Unit/Training/CtsRichEditorNotesTest.php
```diff
@@ -64,7 +64,7 @@ public function it_strips_disallowed_tags_and_attributes_when_saving(): void
 
         $saved = $this->ctsRichContentNotesForCts($html);
 
-        $this->assertSame('<p>Hello link</p>', $saved);
```
CLC0609, line 67: @AxonC I've had to allow links to get through sanitization to allow us to use links in reports, I know you initially stripped them all out for security. Is how I've implemented it okay by you? It will still strip out all scripts.

AxonC (approved):
Id be happy to drop 'CTS' from the title of the rich editor but otherwise LGTM

tests/Unit/Training/CtsRichEditorNotesTest.php
AxonC: That's fine for me

CLC0609:
Renamed to `TrainingRichEditor`

PR #4839, opened 2026-05-28
Auto detect mentoring session adjacent atc

kristiankunc:
Both on the conduct and on the standard report view page.
Included a lot of tests as I did not trust the logic, especially when using datetime

image


AxonC (approved):
We could add a disclaimer explaining this but really happy with this

kristiankunc:
> We could add a disclaimer explaining this but really happy with this

meh, they'll figure it out. Shouldn't be that complicated

PR #4853, opened 2026-05-30
fix: use new View Mentoring Report page instead of CTS on view training place tables

CLC0609:
# Summary of changes
- use new View Mentoring Report page instead of CTS on view training place tables

# Screenshots (if necessary)
image


kristiankunc (approved):
yes please

PR #4858, opened 2026-06-01
fix: stop casting discord id's to ints

CLC0609:
# Summary of changes
- stop casting discord id's to ints

# Screenshots (if necessary)

### DIFF
--- a/app/Libraries/Discord.php
+++ b/app/Libraries/Discord.php
@@ -405,7 +405,7 @@ class Discord
         ];
     }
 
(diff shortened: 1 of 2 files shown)


app/Libraries/Discord.php
kristiankunc: wrong return type

CLC0609: fixed.

kristiankunc (approved):
let's see if this does the trick

PR #4862, opened 2026-06-02
PHPUnit notices

kristiankunc:


CLC0609 (approved):

kristiankunc:
The flag does not work when running in paralell - I'll have to check again

AxonC:
> The flag does not work when running in paralell - I'll have to check again

muchos sados

MrAdder (commented):
There is a way to get this to work with Parallel let me see what I can do

PR #4884, opened 2026-06-04
feat: add email settings to core

CLC0609:
# Summary of changes
Replicates the email settings logic of the cts on core.

- Uses `mship_email_settings` table to store preferences (only stores where users have opted out of an email)
- `EmailType` enum holds all email types that you can opt out of
- Notifications can be assigned a `EmailType`, which will then make them automatically follow the users preferences
- Notifications without an email type are always sent

# Screenshots (if necessary)
image
image

### DIFF
--- a/app/Listeners/Mship/CheckEmailPreferences.php
+++ b/app/Listeners/Mship/CheckEmailPreferences.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace App\Listeners\Mship;
+
+use App\Notifications\Contracts\HasEmailType;
+use Illuminate\Notifications\Events\NotificationSending;
+use Illuminate\Support\Facades\Log;
+
+class CheckEmailPreferences
+{
+    public function handle(NotificationSending $event): bool
+    {
+        $notification = $event->notification;
+        $notifiable = $event->notifiable;
+
+        if (! ($notification instanceof HasEmailType)) {
+            return true;
+        }
+
+        if (! method_exists($notifiable, 'isEmailEnabled')) {
+            return true;
+        }
+
+        $emailType = $notification->getEmailType();
+
+        if (! $notifiable->isEmailEnabled($emailType)) {
+            $notifiableId = method_exists($notifiable, 'getKey') ? $notifiable->getKey() : '?';
+            Log::info("Email suppressed for account {$notifiableId}: {$emailType->value} ({$emailType->label()})");
+
+            return false;
+        }
+
+        return true;
+    }
+}
--- a/app/Models/Mship/Account/EmailSetting.php
+++ b/app/Models/Mship/Account/EmailSetting.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace App\Models\Mship\Account;
+
+use App\Enums\EmailType;
+use App\Models\Mship\Account;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+class EmailSetting extends Model
+{
+    protected $table = 'mship_email_settings';
+
+    protected $fillable = [
+        'account_id',
+        'email_type',
+        'enabled',
+    ];
+
+    protected $casts = [
+        'enabled' => 'boolean',
+    ];
+
+    public function account(): BelongsTo
+    {
+        return $this->belongsTo(Account::class, 'account_id');
+    }
+
+    public function scopeEnabled($query)
+    {
+        return $query->where('enabled', true);
+    }
+
+    public function scopeForType($query, EmailType $type)
+    {
+        return $query->where('email_type', $type->value);
+    }
+}
--- a/app/Models/Mship/Account/EmailSetting.php
+++ b/app/Models/Mship/Account/EmailSetting.php
@@ -2,7 +2,6 @@
 
 namespace App\Models\Mship\Account;
 
-use App\Enums\EmailType;
 use App\Models\Model;
 use App\Models\Mship\Account;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
@@ -24,15 +23,5 @@ class EmailSetting extends Model
     public function account(): BelongsTo
     {
         return $this->belongsTo(Account::class, 'account_id');
-    }
-
-    public function scopeEnabled($query)
-    {
-        return $query->where('enabled', true);
-    }
-
-    public function scopeForType($query, EmailType $type)
-    {
-        return $query->where('email_type', $type->value);
     }
 }
--- a/app/Models/Mship/Account/EmailSetting.php
+++ b/app/Models/Mship/Account/EmailSetting.php
@@ -3,8 +3,8 @@
 namespace App\Models\Mship\Account;
 
 use App\Enums\EmailType;
+use App\Models\Model;
 use App\Models\Mship\Account;
-use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 
 class EmailSetting extends Model
--- a/database/migrations/2026_06_04_000001_create_email_settings_table.php
+++ b/database/migrations/2026_06_04_000001_create_email_settings_table.php
@@ -22,6 +22,6 @@ return new class extends Migration
 
     public function down(): void
     {
-        Schema::dropIfExists('email_settings');
+        Schema::dropIfExists('mship_email_settings');
     }
 };
--- a/database/migrations/2026_06_04_000001_create_email_settings_table.php
+++ b/database/migrations/2026_06_04_000001_create_email_settings_table.php
@@ -1,0 +1,27 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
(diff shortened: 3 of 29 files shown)


kristiankunc (commented):
Do we wanna migrate the legacy email settings from CTS?

database/migrations/2026_06_04_000002_migrate_cts_email_settings_to_core.php
```diff
@@ -0,0 +1,91 @@
...
+        $genMapping = [
+            'exam_conf' => 'examiner_exam_accepted',
+            'exam_cancel' => 'examiner_exam_cancelled',
+        ];
+
+        $rows = [];
+
+        $rtsRows = $cts->table('email_settings_rts')
+            ->join('members', 'members.id', '=', 'email_settings_rts.member_id')
+            ->select('members.cid as account_id', 'email_settings_rts.*')
```
kristiankunc: you actually learned your lesson haha

database/migrations/2026_06_04_000001_create_email_settings_table.php
```diff
@@ -0,0 +1,27 @@
...
+
+            $table->unique(['account_id', 'email_type']);
+            $table->foreign('account_id')->references('id')->on('mship_account')->onDelete('cascade');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::dropIfExists('email_settings');
+    }
```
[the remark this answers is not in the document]
CLC0609: wow.

resources/views/filament/training/pages/email-settings.blade.php
```diff
@@ -0,0 +1,11 @@
+<x-filament-panels::page>
+	<form wire:submit="save">
+		{{ $this->form }}
+
+		<div class="mt-6 flex justify-end">
+			<x-filament::button type="submit">
+				Save Settings
+			</x-filament::button>
+		</div>
+	</form>
```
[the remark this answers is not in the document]
CLC0609: Surely prek would have failed this?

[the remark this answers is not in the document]
CLC0609: Oh prek literally changed it to this 817a648337eb1f7c85df4dd7956abfe697d58b93

database/migrations/2026_06_04_000002_migrate_cts_email_settings_to_core.php
```diff
@@ -0,0 +1,91 @@
...
+                    ];
+                }
+            }
+        }
+
+        // Chunk insert to avoid memory issues
+        $chunks = array_chunk($rows, 500);
+
+        foreach ($chunks as $chunk) {
+            DB::table('mship_email_settings')->insertOrIgnore($chunk);
```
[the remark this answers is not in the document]
kristiankunc: good idea - the table might be large

database/migrations/2026_06_04_000002_migrate_cts_email_settings_to_core.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        Log::info('CTS email settings migrated, '.count($rows).' preferences disabled.');
+    }
+};
```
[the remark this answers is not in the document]
kristiankunc: fine with not implementing a rollback

resources/views/filament/training/pages/email-settings.blade.php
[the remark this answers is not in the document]
kristiankunc: yeah just hallucinations - trust prek

app/Listeners/Mship/CheckEmailPreferences.php
```diff
@@ -0,0 +1,35 @@
...
+class CheckEmailPreferences
+{
+    public function handle(NotificationSending $event): bool
+    {
+        $notification = $event->notification;
+        $notifiable = $event->notifiable;
+
+        if (! ($notification instanceof HasEmailType)) {
+            return true;
+        }
```
AxonC, line 18: I _think_ I really like this. It means we don't have to check each time we dispatch a notification whether or not the preferences are met. Did you note any drawbacks with this approach?

database/migrations/2026_06_04_000001_create_email_settings_table.php
```diff
@@ -0,0 +1,27 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::create('mship_email_settings', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedInteger('account_id');
+            $table->string('email_type', 100);
```
AxonC, line 14: I would be tempted to remove the string length restriction here just incase we need it longer in future. Or at least make it a bit bigger

database/migrations/2026_06_04_000002_migrate_cts_email_settings_to_core.php
```diff
@@ -0,0 +1,86 @@
...
+            if (empty($buffer)) {
+                return;
+            }
+
+            DB::table('mship_email_settings')->insertOrIgnore($buffer);
+            $total += count($buffer);
+            $buffer = [];
+        };
+
+        $process = function (iterable $rows, array $mapping) use (&$buffer, &$total, $chunkSize, $flush) {
```
AxonC, line 40: I'm pretty sure reading it you've done it right, but just double checking for CTS ID and Core ID mismatching :D

database/migrations/2026_06_04_000001_create_email_settings_table.php
```diff
@@ -0,0 +1,27 @@
...
+            $table->boolean('enabled')->default(true);
+            $table->timestamps();
+
+            $table->unique(['account_id', 'email_type']);
```
AxonC, line 18: I"m guessing unique from memory also creates an index to make this query even faster.

database/migrations/2026_06_04_000001_create_email_settings_table.php
```diff
@@ -0,0 +1,27 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::create('mship_email_settings', function (Blueprint $table) {
+            $table->id();
```
AxonC, line 12: Is it superfolous to have an ID field here when the real primary-key identifier is a compound key of account_id and email_type? Or does that just make it more complex?

app/Listeners/Mship/CheckEmailPreferences.php
```diff
@@ -0,0 +1,35 @@
...
+
+        if (! method_exists($notifiable, 'isEmailEnabled')) {
+            return true;
```
AxonC, line 21: Is the return value here what blocks the dispatch? I'm just trying to track the call stack as to what powers the notification being stopped

app/Models/Mship/Account/EmailSetting.php
```diff
@@ -0,0 +1,38 @@
...
+
+    public function scopeEnabled($query)
+    {
+        return $query->where('enabled', true);
+    }
+
+    public function scopeForType($query, EmailType $type)
+    {
+        return $query->where('email_type', $type->value);
+    }
```
AxonC: Would we ever need to query by just type or enabled, or would we always want to query by account, or account and type?

app/Listeners/Mship/CheckEmailPreferences.php
CLC0609: I can't think of many tbh. It's slightly extra work when we create notifications now that we have to create a type and assign it, but if you just don't give it a type it will just always send. So it's not massively extra.

database/migrations/2026_06_04_000001_create_email_settings_table.php
CLC0609: Removed

database/migrations/2026_06_04_000002_migrate_cts_email_settings_to_core.php
CLC0609: Couldn't find it myself for a second there 😆 

`->select('members.cid as account_id', 'email_settings_rts.*')` uses the cid

app/Listeners/Mship/CheckEmailPreferences.php
CLC0609: Yep it's the return that stops it sending

database/migrations/2026_06_04_000001_create_email_settings_table.php
CLC0609: It does make it slightly more complex, but can do if that's something that's wanted. I will say, because we are only storing "opt-out" records, and not every preference, that I don't expect a huge number of rows.

app/Models/Mship/Account/EmailSetting.php
CLC0609: Looking back I don't think I've actually used either of them, just thought I might have when writing the model. Will remove them both.

AxonC (approved):

PR #4887, opened 2026-06-06
feat: allow retention checks to be enabled on any waitinglist

CLC0609:
# Summary of changes
- Allow retention checks to be enabled on any waiting list
- Adds is_vt to waiting list
- Adds specific copy text for VT retention checks

# Screenshots (if necessary)
image
image


kristiankunc:
I'm not sure the backend is really set up for it - have you checked/tested?

CLC0609:
Yep, it was built for multiple to be enabled by the looks of it:
`CreateWaitingListRetentionChecks` runs at 07:00 every day, and checks every waiting list that has retention checks enabled. Loops through all members, if they need a check, `SendWaitingListRetentionCheck` is dispatched, sending the retention check.

The only other thing that might need changing is the copy text if Will wants different

database/migrations/2026_06_12_143453_add_is_visiting_to_training_waiting_list_table.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('training_waiting_list', function (Blueprint $table) {
+            $table->boolean('is_vt')->default(false)->after('department');
```
AxonC: Did you consider adding this into the existing set of feature flags instead of adding a new column?

CLC0609: Updated to use a feature flag instead

AxonC (approved):

PR #4888, opened 2026-06-06
Add discord honeypot

kristiankunc:
Will require a new service to run the `discord:listen` command constantly

TODO: set channel ID

## Features
 - Honeypot channel
 - Times user out for a week
 - Adds a note on core
 - Deletes all messages from the last hour


app/Libraries/Discord.php
```diff
@@ -265,6 +265,183 @@ public function createThreadFromMessage(string $channelId, string $messageId, ar
...
+        }
+
+        // for each text channel, fetch recent messages and filter by author
+        $textChannelTypes = [0, 5]; // GUILD_TEXT, GUILD_NEWS
+        foreach ($channelsResponse->json() as $channel) {
+            if (! in_array($channel['type'] ?? -1, $textChannelTypes, true)) {
+                continue;
+            }
+
+            $response = $this->rateLimitedRequest(
```
CLC0609: We are going to get heavily rate limited by this aren't we? It's going to need to check like 200 channels. 

Could we use `/guilds/{guild.id}/messages/search` ([docs](https://docs.discord.com/developers/resources/message#search-guild-messages)) with a `author_id` instead?

kristiankunc: That is heavily cached and delayed - when we want to especially target very recent messages it proved essentially useless from testing

CLC0609 (approved):
Looks great! Going to put us mods out of a job 😆

AxonC:
How often do you need this to run on the scheduler? Every minute?

kristiankunc:
this is designed to run constantly. it establishes a ws gateway and listens for discord events

PR #4900, opened 2026-06-14
feat: add student to roster if OBS exam is overridden to a pass

CLC0609:
# Summary of changes
- add student to roster if OBS exam is overridden to a pass

# Screenshots (if necessary)


AxonC (approved):
Holy edge-case batman!

PR #4904, opened 2026-06-15
fix: add slight delay to TS assign and remove jobs

CLC0609:
# Summary of changes
- Adds slight delay to try and stop the jobs from overlapping with eachother

# Screenshots (if necessary)


kristiankunc (approved):
reasonable

PR #4924, opened 2026-06-20
Add ICS attachment to booking emails

kristiankunc:
Adds an ICS attachment (https://en.wikipedia.org/wiki/ICalendar)  to booking emails which allows easy "add to calendar" buttons in most mail clients

image


CLC0609 (commented):
Error when accepting a mentoring session
```
Could not parse '2026-06-24 00:00:00 18:00': Failed to parse time string (2026-06-24 00:00:00 18:00) at position 20 (1): Double time specification
```

kristiankunc:
Ah damnit there's the 00-00 time by default, I'll fix that

CLC0609 (approved):

PR #4927, opened 2026-06-21
fix: add null handling to edit action on view account page

CLC0609:
Fixes [bugsnag error](https://app.bugsnag.com/vatsim-uk/core-1/errors/6a36f25540552c86acc228e8?filters[error.status]=new&filters[event.since]=30d)

# Summary of changes
- adds null handling to edit action on view account page

# Screenshots (if necessary)


kristiankunc:
This is odd - how can user be null on the panel?
Are we sure this is not just some weird case?

CLC0609:
Reloading the page after session has expired, hence not logged in. maybe?

CLC0609:
I mean looking at the bugsnag report the user wasn't logged in

kristiankunc (approved):

PR #4928, opened 2026-06-21
fix: change `sesh_id` on `cancel_reason` to be a small_int

CLC0609:
Fixes [bugsnag error](https://app.bugsnag.com/vatsim-uk/core-1/errors/6a37e6b190a602cb082afb94?filters[error.status]=new&filters[event.since]=30d)

# Summary of changes
- Session.id is an int, sesh_id on cancel reason is a small int causing an overflow

# Screenshots (if necessary)


kristiankunc:
What issue does this solve?

edit: Oh sorry am blind

database/migrations/2026_06_21_191145_alter_cancel_reason_table_change_sesh_id_to_int.php
```diff
@@ -0,0 +1,22 @@
...
+            $table->integer('sesh_id')->unsigned()->change();
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::connection('cts')->table('cancel_reason', function (Blueprint $table) {
+            $table->smallInteger('sesh_id')->unsigned()->change();
+        });
+    }
```
AxonC: I can do this manually if it means we have one less migration, just means we have to update the mock

CLC0609: Sounds good, I've removed the migration, and just updated the mock

AxonC (approved):

PR #4941, opened 2026-06-22
Discord tags

kristiankunc:
Discord tag commands for quick info managable via core

image

### DIFF
--- a/app/Console/Commands/Discord/RunDiscordBot.php
+++ b/app/Console/Commands/Discord/RunDiscordBot.php
@@ -121,7 +121,8 @@ class RunDiscordBot extends Command
 
             if (! $tag) {
                 $interaction->respondWithMessage(
-                    MessageBuilder::new()->setContent("Tag `{$key}` not found.")
+                    MessageBuilder::new()->setContent("Tag `{$key}` not found."),
+                    ephemeral: true
                 );
 
                 return;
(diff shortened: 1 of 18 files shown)


app/Console/Commands/Discord/RunDiscordBot.php
```diff
@@ -71,6 +76,74 @@ public function handle()
...
+                );
+
+                return;
+            }
+
+            $tag = DiscordTag::where('key', $key)->first();
+
+            if (! $tag) {
+                $interaction->respondWithMessage(
+                    MessageBuilder::new()->setContent("Tag `{$key}` not found.")
```
CLC0609: Could we make this a ephemeral message so it's only visible to the user that ran the command, just thinking about clogging channels

app/Libraries/Discord.php
```diff
@@ -407,6 +408,133 @@ public function editMessage(string $channelId, string $messageId, array $newCont
...
+            Log::warning('Too many tags for slash command choices, truncating to the first 25');
+            $choices = array_slice($choices, 0, 25);
+        }
+
+        $commandPayload = [
+            'name' => 'tag',
+            'description' => 'Display a tag',
+            'options' => [
+                [
+                    'type' => 3, // STRING
```
CLC0609, line 444: string?

CLC0609 (commented):
Looks great!

app/Libraries/Discord.php
kristiankunc: image

CLC0609: ahhh I thought it was a comment to change the type into a string or something lol

CLC0609 (approved):

PR #4946, opened 2026-06-24
Update s2-syllabus-body.blade.php to include missing topics.

DuckWhsiperer:
Update Basic Training lesson plan 2 to include missing topics.

# Summary of changes

Adds Training AD Overview and Sequencing to lesson two plan.


kristiankunc:
image

Funily enough when scanning thru it yesteray I noticed some of the tables are a bit broken (likely due to my mistakes) - mind fixing those while you're there?

PR #4953, opened 2026-06-25
fix: set minimum availability slot length to 45 minutes

CLC0609:
# Summary of changes
- Do not allow slots < 45 minutes to be created

# Screenshots (if necessary)


app/Filament/Training/Pages/MyTraining/MyAvailability.php
```diff
@@ -153,6 +153,18 @@ public function create(): void
         $startDate = Carbon::parse($data['date_range']['start']);
         $endDate = Carbon::parse($data['date_range']['end']);
 
+        $startTime = Carbon::parse("{$startDate->toDateString()} {$data['from']}", 'UTC');
+        $endTime = Carbon::parse("{$startDate->toDateString()} {$data['to']}", 'UTC');
+
+        if ($startTime->diffInMinutes($endTime) < AvailabilityService::MINIMUM_SLOT_DURATION_MINUTES) {
```
kristiankunc: Can you not use the isValidSlot method here instead? This feels very duplicated

CLC0609: Moved it into its own function, called from both places.

kristiankunc (approved):

PR #4959, opened 2026-06-27
Stop truncating badges on session calendars

kristiankunc:
Truncates name and ensures badge is intact


CLC0609 (changes requested):
Image

Letters expand beyond the badge, could we make sure the text stays within it.

AxonC (approved):

CLC0609 (approved):
Image

Looks great!

PR #4961, opened 2026-06-28
Move endorsements to training panel

kristiankunc:


CLC0609 (commented):
Could we put them into an endorsements navigation category, and then put it below Mentoring maybe? The mentoring section is looking super cluttered like this.

Image

kristiankunc:
image

CLC0609 (approved):

PR #4965, opened 2026-07-01
Add UKCP positions syncing

kristiankunc:
_Waiting for BW to review UKCP positions_

Adds a scheduled sync (daily at 04:00) that pulls controller positions from the UKCP API into the core positions table. New positions are created, existing ones have frequency/ID updated, and positions removed from UKCP are soft-deleted. Includes ukcp_position_id column, soft deletes on the model, a helper to infer position type from callsign, and removes the unused sub_station column. Supports --dry-run for preview.


app/Filament/Admin/Resources/Positions/PositionResource.php
CLC0609, line 35: We should probably disable most of these fields, any changes would just be overridden by UKCP every morning anyways. Only one I think we need enabled would be the position group one.

CLC0609 (commented):
Other than that all looks good!

app/Filament/Admin/Resources/Positions/PositionResource.php
kristiankunc: Good spot!

kristiankunc: Disabled on synced positions + added a link button to UKCP to quickly edit source

CLC0609 (approved):

app/Models/Atc/Position.php
```diff
@@ -110,6 +112,39 @@ public function scopeTemporarilyEndorsable(Builder $query): Builder
...
+    {
+        $suffix = strtoupper(Arr::last(explode('_', $callsign)));
+
+        return match ($suffix) {
+            'ATIS' => self::TYPE_ATIS,
+            'DEL', 'DELIVERY' => self::TYPE_DELIVERY,
+            'GND', 'GROUND' => self::TYPE_GROUND,
+            'TWR', 'TOWER' => self::TYPE_TOWER,
+            'APP', 'APPROACH', 'F_APP', 'N_APP', 'S_APP', 'A_APP',
+            'L_APP', 'D_APP', 'R_APP', 'P_APP', 'Z_APP' => self::TYPE_APPROACH,
```
CLC0609: Missed this on my review, but we probably don't need all the `*_APP` callsigns as the split would only grab the last part anyways right?

PR #4972, opened 2026-07-05
fix: improve search styling on roster page

CLC0609:
# Summary of changes
- improve search styling on roster page

# Screenshots (if necessary)
Before:
image

After:
image


AxonC (approved):
I h8 front-end

PR #5004, opened 2026-07-16
feat: Allow staff to edit an applications facility

WillShaw1100:
Per title, allows an authorised member to change a vt applications facility from, for example C1 Visit to C1 Transfer without the member needing to open a new application

image

image

Will need the 'vt.application.modify.*' permission awarded once in prod

### DIFF
--- a/app/Filament/Admin/Resources/VisitTransfer/VisitTransferApplications/Pages/ViewVisitTransferApplication.php
+++ b/app/Filament/Admin/Resources/VisitTransfer/VisitTransferApplications/Pages/ViewVisitTransferApplication.php
@@ -4,6 +4,7 @@ namespace App\Filament\Admin\Resources\VisitTransfer\VisitTransferApplications\P
 
 use App\Enums\VTCheckStatus;
 use App\Filament\Admin\Resources\VisitTransfer\VisitTransferApplications\VisitTransferApplicationResource;
+use App\Models\Mship\Note\Type;
 use Carbon\Carbon;
 use Filament\Actions\Action;
 use Filament\Forms\Components\Checkbox;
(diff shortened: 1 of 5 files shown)


WillShaw1100:
Ill make a test first

app/Filament/Admin/Resources/VisitTransfer/VisitTransferApplications/Pages/ViewVisitTransferApplication.php
```diff
@@ -147,7 +148,44 @@ public function infolist(Schema $schema): Schema
...
                                     TextEntry::make('pilot_rating')->label('Pilot Rating')->getStateUsing(fn ($record) => ($record->account?->qualification_pilot?->name_long ?? 'Unknown Pilot')),
                                     TextEntry::make('type_string')->label('Facility Type'),
-                                    TextEntry::make('facility.name')->label('Facility Name'),
+                                    TextEntry::make('facility.name')->label('Facility Name')->hintAction(Action::make('change_facility')
+                                        ->label('Edit')
+                                        ->icon('heroicon-m-pencil-square')
+                                        ->color('warning')
+                                        ->modalHeading(fn ($record) => "Change Facility for Application #{$record->public_id}")
+                                        ->modalDescription('This will change the facility for the application. It bypasses facility restrictions including rating requirements.')
```
CLC0609, line 156: It also does not change the checks on the Application record, so if they were disabled on the first facility they will remain disabled even if they should be enabled on the 2nd facility. If your happy with this logic then that's fine by me.

WillShaw1100: Not sure it matters, this will have limited use case for us to switch people from transferring to visiting etc.

PR #5015, opened 2026-07-23
fix: ignore pre 2020 pilot exams

CLC0609:
# Summary of changes
- ignore pre 2020 pilot exams

# Screenshots (if necessary)

### DIFF
--- a/app/Repositories/Cts/ExamResultRepository.php
+++ b/app/Repositories/Cts/ExamResultRepository.php
@@ -18,10 +18,11 @@ class ExamResultRepository
             ->get();
     }
 
-    public function getPassedExamsOfType(string $type): Collection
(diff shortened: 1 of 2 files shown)


app/Repositories/Cts/ExamResultRepository.php
```diff
@@ -18,10 +18,11 @@ public function getRecentPassedExamsOfType(string $type, int $daysConsideredRece
             ->get();
     }
 
-    public function getPassedExamsOfType(string $type): Collection
+    public function getPassedExamsOfType(string $type, ?\DateTimeInterface $since = null): Collection
```
AxonC:
```suggestion
    public function getPassedExamsOfType(string $type, ?Carbon\Carbon $since = null): Collection
```

AxonC (approved):

PR #5017, opened 2026-07-23
fix: allow tables to be shown on mentoring reports

CLC0609:
# Summary of changes
- Allows tables to be shown

# Screenshots (if necessary)
image


AxonC (approved):
Who knew people uses tables in mentoring reports?!!?

PR #5025, opened 2026-07-25
Fix: VT - Change accessors used in Staff Facility Edit

WillShaw1100:
Per title, is accepted should be used instead, this was a mistake on my part.


AxonC:
Can you cover this with a unit test to avoid such a bug in future?

AxonC (approved):

kristiankunc (approved):

PR #5039, opened 2026-07-27
fix: exams test not working after 2200

WillShaw1100:
It was putting the exam over two days causing issues.


kristiankunc (approved):
@CLC0609 sorry but this is hilarious haha

PR #4967, opened 2026-07-01
New booking schema

kristiankunc:
### **User description**
New bookings table in core
 - Optionally links to additional sources (exams, sessions, ...)
 - Optionally links to _now legacy_ cts booking for later migrations
 - Core logic has been migrated to use this table

___

### **PR Type**
Enhancement

___

### **Description**
- Introduce new core Booking model and migration

- Add BookingService with overlap validation and CRUD

- Migrate legacy CTS bookings to new schema

- Update exam listener and repositories to use new Booking model

___

### Diagram Walkthrough

```mermaid
flowchart LR
    A["Booking Model & Schema"] --> B["BookingService\n(CRUD + Overlap)"]
    A --> C["Migration from CTS"]
    D["Exam Listener"] --> A
    E["BookingRepository Refactor"] --> A
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>8 files</summary><table>
<tr>
  <td><strong>CreateCtsBookingEntry.php</strong><dd><code>Refactor exam listener to use new Booking model</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/VATSIM-UK/core/pull/4967/files#diff-652f1d4d25883a19740b862e0001e4348cfd0c030ae23a5befa14cc9eaeb8647">+15/-17</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>Position.php</strong><dd><code>Add bookings relationship to Position model</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/VATSIM-UK/core/pull/4967/files#diff-3d6bca3ba9e951cf3f9a6a8ed3e0c99f54a6fe954d9e6ba34660c13ca4e52029">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>Booking.php</strong><dd><code>Introduce new Booking model with polymorphic relationships</code></dd></td>
[...]

PR #5040, opened 2026-07-28
Stop CTS CID cheating on tests

kristiankunc:
Ensures CTS user ID is never the same as the CID

@CLC0609 relevant files are:
 - `tests/Unit/Training/TrainingPlace/ExamForwardingServiceTest.php`
 - `tests/Feature/TrainingPanel/BaseTrainingPanelTestCase.php`
 - `database/factories/Cts/ExamBookingFactory.php`
 - `database/factories/Cts/MemberFactory.php`

I think these 2 files had an actual bug:
 - `app/Listeners/Training/Exams/CreateCtsBookingEntry.php`
 - `app/Services/Training/CancelPendingExamService.php`
 
 Other diffs are just using the new factory setup

### DIFF
--- a/app/Listeners/Training/Exams/CreateCtsBookingEntry.php
+++ b/app/Listeners/Training/Exams/CreateCtsBookingEntry.php
@@ -20,7 +20,7 @@ class CreateCtsBookingEntry
 
         app(BookingService::class)->create([
             'position_id' => $position?->id,
-            'member_id' => $examBooking->student_id,
+            'member_id' => $examBooking->student?->cid,
             'type' => Booking::TYPE_EXAM,
             'starts_at' => Carbon::parse($examBooking->taken_date)->format('Y-m-d').' '.$examBooking->taken_from,
             'ends_at' => Carbon::parse($examBooking->taken_date)->format('Y-m-d').' '.$examBooking->taken_to,
--- a/database/factories/Cts/MemberFactory.php
+++ b/database/factories/Cts/MemberFactory.php
@@ -3,6 +3,7 @@
 namespace Database\Factories\Cts;
 
 use App\Models\Cts\Member;
+use App\Models\Mship\Account;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Factories\Factory;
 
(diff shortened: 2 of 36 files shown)


app/Listeners/Training/Exams/CreateCtsBookingEntry.php
```diff
@@ -20,7 +20,7 @@ public function handle(ExamAccepted $event): void
 
         app(BookingService::class)->create([
             'position_id' => $position?->id,
-            'member_id' => $examBooking->student_id,
+            'member_id' => $examBooking->student?->cid,
```
CLC0609, line 23: Shouldn't this be cts id?

database/factories/Cts/MemberFactory.php
```diff
@@ -3,24 +3,41 @@
...
+
+    /**
+     * Link this member to a real account via `cid`, while keeping `id` deliberately
+     * different so tests exercise the `cid` translation instead of assuming id === cid.
+     */
+    public function forAccount(Account $account): static
+    {
+        return $this->state(fn () => [
+            'cid' => $account->id,
+            'id' => $account->id + 5000000,
```
CLC0609, line 40: We have
```
public function generateCTSInternalID(int $cid): int // NOSONAR
    {

        if ($cid < 800000) {
            return $cid + 2000000;
        } elseif ($cid < 1323000) {
            return $cid;
        } elseif ($cid < 1800000) {
            return $cid - 1000000;
        } else {
            return $cid + 2000000;
        }
    }
``` 
in `HasCTSAccount` which we could probably use here. Not that it really matters.

app/Listeners/Training/Exams/CreateCtsBookingEntry.php
kristiankunc: I think not anymore since the new bookings table / modification of the booking setup (https://github.com/VATSIM-UK/core/pull/4967)

database/factories/Cts/MemberFactory.php
kristiankunc: Imo not, this is the code on CTS which we attempt to mirror on core. For certain CIDs it produces `id = cid` which is going to result in flaky tests if for some cases. Imo this method should not be used as there is still a chance of this happening. We should ALWAYS assume the worst (`cid != id`) that way tests are forced to use the proper relationship without there being a chance of this cheating and a lucky test success

kristiankunc:
wtf

CLC0609 (approved):

PR #5051, opened 2026-07-31
Rich text editor links show open in new tab by default

kristiankunc:


CLC0609 (commented):
Image
Not defaulting for me

CLC0609 (commented):
still nothing for me

kristiankunc:
Should be good now - tooks some more tweaking than expected

CLC0609 (approved):
works like a charm

PR #5060, opened 2026-08-01
Add full opencode mcp tooling

kristiankunc:
Bunch of useful opencode MCPs, may take a look at some skills but I still regard those as somewhat useless


redrails:
May I suggest [grill-me](https://github.com/mattpocock/skills/tree/main/skills/productivity/grill-me) - It's one I've used extensively and works quite well in plan mode for architectural context and normally produces very accurate plans.

kristiankunc:
Those skills are mostly personal preference so I probably won't be adding those

PR #5063, opened 2026-08-01
fix: ensure only "certified" members on gander roster are synced

CLC0609:
# Summary of changes
- ensure only "certified" members on gander roster are synced

# Screenshots (if necessary)


kristiankunc (approved):
lgtm

PR #5080, opened 2026-08-03
Style: Update site Navigation + remove mention of the forums

WillShaw1100:
Fix a few links and remove reference to the forums except in policy


resources/views/components/nav.blade.php
```diff
@@ -291,13 +291,13 @@ class="nav-dropdown-link">Log Out</a>
 								<li><a href="{{ route('site.community.teamspeak') }}" class="nav-dropdown-link">TeamSpeak</a></li>
 								<li><a href="{{ route('mship.manage.dashboard') }}" class="nav-dropdown-link">Discord</a></li>
 								<li class="nav-dropdown-divider"></li>
-								<li><a href="https://cts.vatsim.uk/" class="nav-dropdown-link">Training System</a></li>
+								<li><a href="{{ route('filament.training.pages.dashboard') }}" class="nav-dropdown-link">Training System</a>
```
CLC0609: Personally I would leave this as CTS for now. Training is still in beta so we still want CTS linked. 
Plus we already have a link to the training panel on the navigation bar via the icon.

CLC0609 (changes requested):
Thanks, just one comment from me.

resources/views/components/nav.blade.php
kristiankunc: agreed, lets leave cts for now

WillShaw1100: This was one of the suggestions made after I asked people why they wernt using core. I had several responses saying theyd never heard that there was a training page and others who wernt aware of its features. Ill revert this for now but maybe there needs to be better advertisemwnt in general otherwise itll be a big shock when cts is removed.

CLC0609: There is now a big notice on every CTS page to hopefully increase usage of the training panel.

kristiankunc (approved):

CLC0609 (approved):

PR #5097, opened 2026-08-07
Enforce 15-minute booking slots on the calendar

AxonC:
## Summary
- Reject booking start/end times that are not on 15-minute boundaries in `Calendar::createBooking`
- Keep timeline drag selection snapped/clamped to quarter-hour slots within the day
- Block free typing in the create modal date/time fields and snap any prefilled times onto valid slots

## Test plan
- [ ] Open `/atc/bookings/calendar` and create a booking via the Book button using only the select pickers
- [ ] Confirm date/time fields cannot be typed into (picker / arrow keys only)
- [ ] Drag a slot on a position row and confirm start/end stay on `:00` / `:15` / `:30` / `:45`
- [ ] Attempt to create a booking with off-boundary times via Livewire (or temporarily bypass UI) and confirm a booking-error is returned
- [ ] Run `php artisan test tests/Feature/Bookings/CalendarTest.php --filter=fifteen_minute`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/Livewire/Bookings/Calendar.php
+++ b/app/Livewire/Bookings/Calendar.php
@@ -640,8 +640,8 @@ class Calendar extends Component
                     return;
                 }
 
-                if ($startsAt->equalTo($endsAt)) {
-                    $this->dispatch('booking-error', message: 'Booking length cannot be zero minutes.');
+                if ($endsAt->lessThanOrEqualTo($startsAt)) {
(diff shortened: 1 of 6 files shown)


kristiankunc (commented):
While you're messing with the input dropdwn - maybe hide end times that are earlier or same as start time - bookings cannot end before staring

PR #5098, opened 2026-08-08
chore: show PHP notice details in test output

MrAdder:
## Show PHP notices and deprecations in test output

Right now our test output only gives us the count:

```
Tests: 1622, Assertions: 4231, Deprecations: 7, Notices: 2.
```

Seven deprecations — but *where*? No file, no message, so they get ignored until one becomes a real bug on a PHP upgrade.

This PR adds one attribute to `phpunit.xml`:

```diff
- <phpunit ... stopOnFailure="false" ...>
+ <phpunit ... stopOnFailure="false" displayDetailsOnAllIssues="true" ...>
```

Now we get file, line, and message for each one.

**We keep parallel runs.** That was the original worry, and it was unfounded — ParaTest collects issues from all workers and prints them in the final summary. Verified locally against our pinned ParaTest 7.20 / PHPUnit 12.5. The CI workflow is untouched: `phpunit.xml` is read by both PHPUnit and ParaTest, so this applies locally *and* in CI with no flags to keep in sync.

**Nothing new fails** — this only changes what gets printed. Expect a longer CI log on the first run; that's the existing backlog becoming visible. If vendor deprecations prove too noisy we can scope to first-party code later via `<source ignoreIndirectDeprecations>`.


MrAdder:
Tests fail due to it using outdated branch used in #4862

kristiankunc:
This looks good, I'll fix it later on, thanks

kristiankunc (approved):
thx for this

PR #5126, opened 2026-08-18
fix: changes to ASRs for Pilot TPs

CLC0609:
# Summary of changes
- Set availability check length to be 5 days
- Ensure no checks are created within 7 days of a session

# Screenshots (if necessary)

### DIFF
--- a/config/training.php
+++ b/config/training.php
@@ -9,7 +9,7 @@ return [
 
     'availability_warning_days' => [
         'atc' => (int) env('TRAINING_AVAILABILITY_WARNING_DAYS_ATC', 5),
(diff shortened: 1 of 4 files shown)


config/training.php
```diff
@@ -9,7 +9,7 @@
 
     'availability_warning_days' => [
         'atc' => (int) env('TRAINING_AVAILABILITY_WARNING_DAYS_ATC', 5),
```
CLC0609, line 11: @~~AxonC isn't this meant to be 7 for ATC TPs?~~ nevermind I got mixed up between PTD and ATC.

kristiankunc (approved):

PR #5133, opened 2026-08-22
feat: new searchable position selector question type

CLC0609:
Depends on #5132

# Summary of changes
- Adds a new position selector question type that will be used on the new ATC form

# Screenshots (if necessary)
image


kristiankunc (approved):
love it

PR #5140, opened 2026-08-24
fix: improve header styling on feedback resource page

CLC0609:
# Summary of changes
- All the information shown is still the same, just displayed in a better way

# Screenshots (if necessary)
Before:
image

After:
image


kristiankunc (approved):
looks much cleaner

PR #5143, opened 2026-08-26
Local test running

kristiankunc:
Fix the always-failing tests on localhost


kristiankunc:
fixed partialy in https://github.com/VATSIM-UK/core/pull/5174 + other devonctainer related PRs

PR #5150, opened 2026-08-28
Add event management to core

kristiankunc:
Complete event management migration from CTS to core

 - Admin panel management
 - Publishing checklist
 - Publishing a calendar booking + on homepage

image


app/Filament/Admin/Resources/Events/EventResource.php
```diff
@@ -0,0 +1,194 @@
...
+                            ->label('Positions')
+                            ->relationship('positions', 'callsign')
+                            ->multiple()
+                            ->preload()
+                            ->searchable()
+                            ->helperText('The ATC positions the event covers.'),
+                        Select::make('manager_id')
+                            ->label('Event manager')
+                            ->relationship('manager', 'name_first')
+                            ->getOptionLabelFromRecordUsing(fn (Account $record): string => "{$record->name_first} {$record->name_last} (CID {$record->id})")
```
CLC0609:
```suggestion
                            ->getOptionLabelFromRecordUsing(fn (Account $record): string => "{$record->name_first} {$record->name_last} ({$record->id})")
```

CLC0609 (commented):
Basic review just from having a click through. Some ideas of QOL Features.

1, I think it looks better as a checkbox list as opposed to a toggle list.
Image
2. I don't think you should be able to check anything on create? I feel the best process would be create first and then the checklist appears on the edit.
3. Description box is weird. You can only edit by clicking on a specific line. I believe we fixed this in training somewhere to make the whole thing clickab;e
4. Could be cool to track who checked each box, so instead of each thing being a boolean store a cid and then show next to the item, once it's selected who did it? if thats needed.
5. Can we scope the Event Manager box to users with `admin.access`
6. Can we make it autosave when a checkbox is marked. I feel the "Save" should only have to be used when you are updating  the actual information regarding the event.
7.  Positions list on table should collapse
Image
8. Checklist column maybe? changing colours based on how many things are checked.
Image
9. Up for discussion on this one. But should "Details" be changable anyways? I feel it'll cause issues if we have checked "Rostered" and published it and then decide to change the name? Maybe only certain information should be changable.
10. Republishing doesn't update the "Published at" date, this should probably also have an account associated with it. "30 Aug 2026 19:01 by 10000001"

kristiankunc:
Think I fixed all - added a new modular table for the checklist items with the tracking as well.

image

image

AxonC (approved):

PR #5160, opened 2026-09-05
Sync bookings to .net bookings calendar

kristiankunc:
We add every booking which has a calsign existing in the `positions` table so that we don't push SB sessions.
Identified by the extra field for each booking, handles all CRUD operations by hooking into the relevant eloquent methods.

### DIFF
--- a/tests/Unit/Services/Bookings/VatsimNetBookingSyncServiceTest.php
+++ b/tests/Unit/Services/Bookings/VatsimNetBookingSyncServiceTest.php
@@ -1,0 +1,214 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Unit\Services\Bookings;
+
+use App\Models\Atc\Position;
+use App\Models\Booking;
+use App\Models\Cts\ExamBooking;
+use App\Models\Cts\Member;
+use App\Models\Cts\PracticalExaminers;
+use App\Models\Cts\Session;
+use App\Models\Mship\Account;
+use App\Services\Bookings\VatsimNetBookingSyncService;
+use Illuminate\Support\Facades\Http;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCase;
+
+class VatsimNetBookingSyncServiceTest extends TestCase
+{
+    private VatsimNetBookingSyncService $service;
+
+    private Account $examiner;
+
+    private Account $mentor;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->service = app(VatsimNetBookingSyncService::class);
+    }
+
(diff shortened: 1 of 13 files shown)


tests/Unit/Services/Bookings/VatsimNetBookingSyncServiceTest.php
```diff
@@ -0,0 +1,214 @@
...
+
+        $booking = $this->makeExamBooking();
+
+        $this->service->sync($booking);
+
+        Http::assertSent(fn ($request) => $request['cid'] === $this->examiner->id && $request['type'] === 'exam');
+    }
+
+    #[Test]
+    public function it_resolves_the_mentor_for_mentoring_bookings(): void
```
AxonC, line 87: How does this compare to our current policy of anonymity for mentors?

tests/Unit/Services/Bookings/VatsimNetBookingSyncServiceTest.php
```diff
@@ -0,0 +1,214 @@
...
+
+        $booking = Booking::factory()->forEvent()->create();
+
+        $this->service->sync($booking);
+
+        Http::assertNothingSent();
+    }
+
+    #[Test]
+    public function it_skips_virtual_positions(): void
```
AxonC, line 111: Nothing PTD related can leak in here can it? Could we add a test for regression purposes?

tests/Unit/Services/Bookings/VatsimNetBookingSyncServiceTest.php
kristiankunc: So mentors should always be hidden? I thought the policy was that we show mentor (and mentor only)

AxonC: If that is what is in bookings I'm happy just wanted to double check

AxonC (approved):

tests/Unit/Services/Bookings/VatsimNetBookingSyncServiceTest.php
kristiankunc: Yes, mentoring/exams show mentor/examiner as the _owner_. That's been taken from the CTS implementation
image

tests/Unit/Services/Bookings/VatsimNetBookingSyncServiceTest.php
kristiankunc: Will add a test but shouldn't - if a position does not exist in the positions table, we never push it (easy sweatbox detection, should work for PTD as well)

PR #5181, opened 2026-09-11
feat: changes to the view mentoring report policy

CLC0609:
# Summary of changes
New rules on who can access or see a mentoring report:
1. The student who attended the session can always see their own report.
2. Holders of  `training.mentoring.view.*`  or  `training.mentoring.reports.view-all`  see everything.
3. Training Group Instructors see reports within their category ladder, no training place requirement.
4. Mentors see reports for categories below the students TP, but only when the student holds an active training place.

# Screenshots (if necessary)


kristiankunc (approved):

.freebuff/project-id
```diff
@@ -0,0 +1 @@
+deaeddda-9576-4e38-8316-bd6971514878
```
AxonC: What is this lovely string doing here?

CLC0609: Oh I really need to figure out how to stop it from doing that 😭😭

CLC0609: removed.

AxonC (approved):

PR #4543, opened 2026-02-21
Move navbar away from Bootstrap to Tailwind

AxonC:
fixes #4847

PR #5189, opened 2026-09-14
Make beta dashboard the default view

kristiankunc:
been in beta for ages, I like it - no issues


CLC0609:
Did we fix the navigation bar being black?

kristiankunc:
That was fixed in https://github.com/VATSIM-UK/core/pull/4543

PR #5192, opened 2026-09-15
feat: automated moodle flags on waiting lists

CLC0609:
# Summary of changes
- Allows a "Moodle Flag" to be created on a waiting list. The user can select a moodle course, and then an exam within that course. When the user passes that quiz, they flag is marked.

Unable to add tests as we don't have a moodle db.

# Screenshots (if necessary)
image
image

image
image


kristiankunc:
Have you verified the moodle version you tested on and what we are actually running?

CLC0609:
Nope. The schemas for the grades and courses haven't changed from what I can see from documentation. But I cannot see the Moodle version or any of the DB for Moodle. Hence why I've asked Calum for a review.

PR #5208, opened 2026-09-20
fix: resolve issues with students not appearing on availability gantt

CLC0609:
# Summary of changes
- Students who previously did not pass an exam should be shown
- Students who had a session conducted earlier in the day should show

# Screenshots (if necessary)


kristiankunc (approved):
I like the tests

PR #5210, opened 2026-09-21
feat: add warning on mentoring page regarding less students showing

CLC0609:
# Summary of changes
- Adds temporary callout

image

# Screenshots (if necessary)


kristiankunc (approved):
should be removed once we get rid of cts altogether

